PR #11, opened 2015-09-26
Fix integer underflow on compare function

hodduc:
On compare(), If 'id' field of two object is different, it returns int(ao.id - bo.id).
This may cause problem, because offset.id is unsigned 32bit integer. If ao.id < bo.id, ao.id - bo.id underflows.

(Same patch on capnproto v1: https://github.com/glycerine/go-capnproto/pull/34)

### DIFF
--- a/capn.go
+++ b/capn.go
@@ -277,7 +277,7 @@ func compare(a, b rbtree.Item) int {
 	ao := a.(offset)
 	bo := b.(offset)
 	if ao.id != bo.id {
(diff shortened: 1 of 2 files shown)


zombiezen:
Nice. Can you add the corresponding unit test to `capn_test.go`?

hodduc:
Sure, I rebased my commit with the corresponding TC.

zombiezen:
Thanks for finding this!

PR #18, opened 2015-12-28
TextList.AtBytes() and StringBytes() allow string field access as []byte and without allocation

### LINKED ISSUE #17: allow returning []byte in place of string fields
On v1, I was seeing alot of allocation due to string copying. So I added 2 small upgrades:

a) BenchmarkUnmarshalCapnpZeroCopyNoAlloc() in mem.go

b) <string field>Bytes() and TextList.AtBytes() methods that return the []byte slice directly instead of copying it to a string.

The reduction in memory allocation means that I can then read a ton of mmap-ed data and analyze the string fields, all while doing *zero-allocation* and having zero GC pressure.

See https://github.com/glycerine/go-capnproto/commit/db36ab24140ac1737830f5a0d8c87ea26f367fbb

comparisons:
~~~
BenchmarkUnmarshalCapnp-4                    2000000           605 ns/op     766.08 MB/s         256 B/op          5 allocs/op              
BenchmarkUnmarshalCapnpZeroCopy-4           10000000           182 ns/op    2539.81 MB/s          88 B/op          3 allocs/op              
BenchmarkUnmarshalCapnpZeroCopyNoAlloc-4    100000000           24.9 ns/op  18668.12 MB/s          0 B/op          0 allocs/op   
~~~
and
~~~
BenchmarkWithoutStringBytes-4               20000000            78.6 ns/op        32 B/op          1 allocs/op   
BenchmarkStringBytes-4                      50000000            29.4 ns/op         0 B/op          0 allocs/op                              
~~~

It would be great to add similar API []byte access to v2.

glycerine:
Fixes #17 

still todo: needs test/benchmarks. I wrote these for v1 here:
https://github.com/glycerine/go-capnproto/blob/master/stringbytes_test.go

### DIFF
--- a/capnpc-go/templates.go
+++ b/capnpc-go/templates.go
@@ -191,6 +191,14 @@ func (s {{.Node.Name}}) {{.Field.Name|title}}() (string, error) {
 	{{end}}
 }
 
+func (s {{.Node.Name}}) {{.Field.Name|title}}Bytes() ([]byte, error) {
+	p, err := s.Struct.Pointer({{.Field.Slot.Offset}})
+	if err != nil {
+		return nil, err
+	}
+	return {{capnp}}.ToData(p), nil
+}
+
 func (s {{.Node.Name}}) Set{{.Field.Name|title}}(v string) error {
 	{{template "settag" .}}
 	t, err := {{capnp}}.NewText(s.Struct.Segment(), v)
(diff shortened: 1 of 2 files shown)


zombiezen:
Thanks Jason! Opened #19 for the benchmarks.

PR #31, opened 2016-05-15
First pass at including the capnproto base schema

zenhack:
`gen.sh` is the only bit of this that's done manually; everything else is generated.

Per the commit message, at some point I'd like to turn this into something than I can reuse for my sandstorm packages without copypasta.

Also, I noticed there's a little bit of logic in the build system already working with a few of the schemas (schema.capnp and rpc.capnp), but wasn't entirely sure how to go about integrating those.


zenhack:
Ack, the test failures are I think because travis is using the stable version of capnproto (and the schema are from git); they look familiar.

Last I looked the holdup on a stable release was c++11 support in microsoft's compilers; if we wait for a new stable it might be a while. Thoughts on how to handle that?

zombiezen:
No, Travis build doesn't bring in the Cap'n Proto compiler anymore, since it really slowed the build to do it hermetically.  The build failures here are because the generated code you've committed has compile errors.  `capnpc-go` could probably be a bit better about not creating bad code (or at least erroring out), but if you have name conflicts, it will just silently generate code that won't compile.

zenhack:
Hmm, whatever I generated in go.sandstorm seemed to work. I'll have to look at that and see what changed.

zenhack:
Looks like I was mistaken about it having worked on my end; just never ended up building schema or persistent.

The failures in "schema" seem to all be name collisions with  the embedded field "Struct." Is there any reason we couldn't give that field an non-exported, explicit name like "structPtr" (obviously can't do "struct")? Am I correct in thinking that picking a non-exported name would prevent collisions with anything we generate? 

For the stuff in persistent, I'm really not sure what's going on there --- the value of that constant doesn't appear in the schema, or anywhere else in the whole source tree.

zenhack:
Ah, I see, the compiler seems to generate constant values for annotations somehow, so that's where it's coming from. In this case the schema had "Persistent" and "persistent," but we uppercase the latter so it will be exported. Probably need to have a fallback plan for that case.

zombiezen:
You should be able to use the `$Go.name` annotation to change the generated name. I use that already for schema.capnp.

zenhack:
OK, I'll probably do that for now. As you say, it would be nice long term it would be nice if the compiler plugin were smart enough to handle this. Maybe someday I'll try tackling that, but for right now it feels like a tad too much yak shaving.

zenhack:
Second pass, let me know what you think.

While I was doing this I got the following error running the rpc package's test suite on dd1345f:

```
2016/05/18 22:46:07 rpc: aborted by remote: rpc: shutdown
2016/05/18 22:46:07 rpc: aborted by remote: rpc: shutdown
panic: question.reject called more than once

goroutine 69 [running]:
panic(0x6186c0, 0xc8200a25a0)
	/usr/lib/go/src/runtime/panic.go:464 +0x3e6
zombiezen.com/go/capnproto2/rpc.(*question).reject(0xc8200d24e0, 0x7f3c549be002, 0x7f3c549be028, 0xc82000a580)
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/question.go:159 +0x10c
zombiezen.com/go/capnproto2/rpc.(*Conn).handleCancel(0xc8200de540, 0xc8200d24e0)
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/rpc.go:360 +0x80
zombiezen.com/go/capnproto2/rpc.(*Conn).coordinate(0xc8200de540)
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/rpc.go:154 +0x5d3
zombiezen.com/go/capnproto2/rpc.(*Conn).(zombiezen.com/go/capnproto2/rpc.coordinate)-fm()
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/rpc.go:100 +0x20
zombiezen.com/go/capnproto2/rpc.(*manager).do.func1(0xc8200de560, 0xc8200a2470)
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/manager.go:50 +0x52
created by zombiezen.com/go/capnproto2/rpc.(*manager).do
	/home/isd/pkg/go/src/zombiezen.com/go/capnproto2/rpc/manager.go:51 +0xb4
exit status 2
FAIL	zombiezen.com/go/capnproto2/rpc	0.011s
```

I haven't reproduced it since, and haven't done a ton of investigation. but the schema is (module comments and namespace annotations) identical. The generated source has this diff:

```
--- capnp/rpc/rpc.capnp.go	2016-05-18 22:54:25.852321725 -0400
+++ rpc/rpccapnp/rpc.capnp.go	2016-05-18 23:01:34.783331340 -0400
@@ -1,4 +1,4 @@
-package rpc
+package rpccapnp
 
 // AUTO GENERATED - DO NOT EDIT
 
@@ -2525,11 +2525,7 @@
 	if err != nil {
 		return nil, err
 	}
-	d := p.Data()
-	if len(d) == 0 {
-		return d, nil
-	}
-	return d[:len(d)-1], nil
+	return p.Data(), nil
 }
 ```
I'm not sure exactly where that's coming from.

It seems *likely* that whatever is causing this is present in master, but I don't want to just assume that. I think we should wait on merging until that's understood, but still interested in feedback in the meantime.

**Comment by zenhack** [comment]
...probably the first thing to investigate is exactly what that diff is about.

**Comment by zombiezen** [comment]
Woah, haven't seen that crash before, but I do suspect there are some race conditions in the `rpc` package. :(  Don't think it's related to your work.

The diff should just be that you have a slightly outdated capnpc-go, just pull and reinstall it.

**Comment by zenhack** [comment]
Your guess about the outdated compiler plugin was correct; I updated it and regenerated; the generated source is now identical (except for the pacckage name). I've rebased, so it's included in this pr now.

Given that the output is effectively identical, I'm comfortable assuming the bug isn't be cause of this pr, so probably no need to hold up merging on that.

Let me know if there's anything in this pr that you think needs tweaking.

**Comment by zenhack** [comment]
@zombiezen, ping?

**Comment by zombiezen** [comment]
From yak-shaving to bikeshedding: Can you call this directory something different? I'm not fond of it colliding with the package name of the root.  Perhaps something like "std"

**Comment by zenhack** [comment]
The reason I named it that is because it makes the include statements in the capnp files work as is (not just in this repo, but also the sandstorm one I'm building). Given the extra maintience burden, I'm inclined to suggest it would be better to just deal with the name collision, but if after considering you still prefer to rename it I'll defer.

**Comment by zombiezen** [comment]
Ah, I see what you're saying. Perhaps "std/capnp" or "stdlib/capnp"?

**Comment by zenhack** [comment]
@zombiezen, do we care about breaking include paths for existing code?

**Comment by zenhack** [comment]
Moved stuff to `std/capnp`, except for the main capnp package. Were you talking about moving just the schema there, or the whole root package? If the former, that gets weird because again you have two packages by the same name.

**Comment by zombiezen** [comment]
Just the schema.  So I'm imagining:

* zombiezen.com/go/capnproto2/
  * normal go files as part of normal package
  * std/
    * go.capnp
    * capnp/
      * all of the upstream schemas...

**Comment by zenhack** [comment]
Did you mean for the generated code to be in std/ or std/capnp/ ? The latter has the disadvantage of a package name collision with the root package.

**Comment by zenhack** [comment]
Or did you mean *just* the schema, for the include path, but keep the generated code in the same package?

**Comment by zenhack** [comment]
@zombiezen, went with the last option and implemented it. Let me know what you think.

File: std/capnp/README.md [hunk 1]
```diff
@@ -0,0 +1,58 @@
...
+
+The schemas checked in to this repository are those in capnproto commit
+ad4079b (master at the time of writing). Unfortunately, the stable
+release is *very* old, and some schemas out in the wild (notably
+sandstorm) have started expecting more recent versions of the base
+schemas.
+
+# Directory Structure
+
+The schema themselves are stored in this directory, rather than in the
```
zombiezen: Update this paragraph, since I think it's out of date.

std/capnp/README.md
```diff
@@ -0,0 +1,58 @@
...
+To address this kind of issue, some of the schema have been manually
+modified after importing, adding `$Go.name` annotations which prevent
+these errors.
+
+# Versions
+
+The schemas checked in to this repository are those in capnproto commit
+ad4079b (master at the time of writing). Unfortunately, the stable
+release is *very* old, and some schemas out in the wild (notably
+sandstorm) have started expecting more recent versions of the base
```
zombiezen: @kentonv Is this going to be fixed soon-ish? When's 0.6 due?

zombiezen:
:+1: This looks great, thanks Ian!

Just a small doc fix, and I think we're good to go.

std/capnp/README.md
kentonv: Basically Sandstorm and Cap'n Proto are developed together and when dealing with Sandstorm you're expected to use Cap'n Proto master. Cap'n Proto releases, in contrast, are expected to be well-tested across many platforms, whereas Sandstorm is linux-exclusive. In particular we struggle to keep the Windows build up-to-date as we don't actually have any Windows machines here.

That said, there has been progress recently by contributors getting the Windows build caught up, after which it may make sense to do a release. Maybe in the next few months?

zombiezen: SGTM. I think it's okay for us to ship closer to master, since the schemas would be backward-and-forward compatible.  If you don't want us to depend on that, then maybe we can have another branch that does.

std/capnp/README.md
zenhack: I don't think it's strictly incorrect (the README got moved with everything else, so "this directory" still refers to the right thing), but I pushed some improvements, since I think this could use them.

PR #35, opened 2016-06-12
Fixed newline issues with generated code

pavius:
* Schemas with structure groups (like below) would generate incorrect code. Compiling this schema:

```
@0x83c2b5818e83ab19;

$Go.package("template_fix");

struct SomeMisguidedStruct {

    someGroup :group {

        someGroupField @0 :UInt64;
    }
}
```

... would result in the following erroneous code being generated (snipped for relevant part):

```
func (s SomeMisguidedStruct) SomeGroup() SomeMisguidedStruct_someGroup { return SomeMisguidedStruct_someGroup(s) }func (s SomeMisguidedStruct_someGroup) SomeGroupField() uint64 {
    return s.Struct.Uint64(0)
}
```

After modification of the structGroup template (and re-generation of `templates.go`), the above is corrected and proper, compiling code is generated.

* Changed templates to inject newlines after struct ReadRoot* and list At (which previously had no newlines between themselves and the following line). This is just a pedantic fix - the generated code compiled fine (as opposed to the above)

Tested manually and made sure that the differences in `templates.go` are only `\n`.


zombiezen:
:+1: Thanks! I've added the snippet you gave as a test case for `capnpc-go`.

pavius:
Cool, at an initial glance I thought this portion wasn't covered by unit tests. Was obviously wrong :P

zombiezen:
They're more integration test-y than anything else.. I have some spot unit tests here and there, but in general I use integration_test.go in the root package to test the generated code functionality, and then the test I modified is useful for checking that generated code is still parseable (usually for Travis builds).

Issue #2, opened 2015-09-20
rpc.Conn doesn't support promise capabilities


zombiezen:
_From @zombiezen on August 6, 2015 21:1_

The `CapDescriptor` struct has a `senderPromise` field which `rpc.Conn` does not support, along with the corresponding `Resolve` message.  The Go implementation does not emit this descriptor type, so it is only a concern if interoperating with another implementation that does.

_Copied from original issue: zombiezen/go-capnproto#2_

zenhack:
As you suspected, (I think) this is causing problems trying to do stuff with WebSession. I'm getting these errors on the console for [irc-idler][1]: 

```
2016/05/23 17:44:18 rpc: unknown capability type senderPromise
sandstorm/supervisor.c++:1893: error: exception = capnp/rpc.c++:2643: disconnected: RpcSystem was destroyed.
stack: 0x44f1f0 0x4616c0 0x46de50 0x5e2fa0 0x5d0560 0x5cfe80
2016/05/23 17:44:18 rpc: aborted by remote: Peer did not implement required RPC message type.; (uint)message.which() = 2
```

[1]: https://github.com/zenhack/irc-idler

zombiezen:
So a partial workaround is to use an error client in `populateMessageCapTable`.  Unfortunately, this will break down if you need to actually use the capability (which I can't recall whether you need it or not).  Does that work?

zombiezen:
The actual fix is a larger semantic change and I don't have the time to work on it right now.

zenhack:
My guess is that's not going to cut it, but I should do some more digging before I let you quote me on that. It's not what I'm hitting right this second, but it looks like streaming HTTP responses involves a promise, so without them I'd have to buffer the whole thing and send when the connection is closed. Haven't looked at the websocket calls, but my guess is that's not going to go well either.

Any pointers about what needs to change for a proper fix?

zombiezen:
Drat. It's been a while since I've looked at the spec, but generally:

* You would need to create a new "wire promise" capability type
* Handle the `Resolve` message in the coordinator goroutine in `handleMessage`
* After a promise has resolved, if it is a capability that the sender has, then you will need to send a disembargo message and enqueue calls in an `embargoClient` until the embargo is lifted.

[...]

PR #38, opened 2016-07-13
Workaround for #2 as discussed on the bug tracker.

zenhack:
The workaround is described in a comment included in the patch.

Please scruitinize this carefuly; I'm not 100% sure I've understood the solution and/or the code. This does get rid of the errors on the sandstorm console I described, though I haven't done more thorough testing.

### DIFF
--- a/rpc/rpc_test.go
+++ b/rpc/rpc_test.go
@@ -78,15 +78,57 @@ func TestBootstrapFulfilled(t *testing.T) {
 	bootstrapAndFulfill(t, clientCtx, conn, p)
 }
 
-func bootstrapAndFulfill(t *testing.T, ctx context.Context, conn *rpc.Conn, p rpc.Transport) capnp.Client {
-	client, bootstrapID := readBootstrap(t, ctx, conn, p)
+// Receive a Finish message for the given question ID.
+//
+// Immediately releases any capabilities in the message.
+//
+// The test fails if any of the following occur:
+//
+// * An error occurs when reading the message
+// * The message is not of type `Finish`
+// * The message's question ID is not equal to `questionID`.
+//
+// Parameters:
+//
+// t: the *testing.T object for the test in which we're running.
+// ctx: The context to be used when sending the message.
+// p: The rpc.Transport to send the message on.
+// questionID: The expected question ID.
+func recvFinish(t *testing.T, ctx context.Context, p rpc.Transport, questionID uint32) {
+	if finish, err := p.RecvMessage(ctx); err != nil {
+		t.Fatal("error reading Finish:", err)
+	} else if finish.Which() != rpccapnp.Message_Which_finish {
+		t.Fatalf("message sent is %v; want Message_Which_finish", finish.Which())
+	} else {
+		f, err := finish.Finish()
+		if err != nil {
+			t.Fatal(err)
+		}
+		if id := f.QuestionId(); id != questionID {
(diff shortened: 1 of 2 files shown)


zombiezen:
This looks right to me.  I'm at Gophercon right now, and I'd like to write a test to exercise this code path before signing off.

zenhack:
Quoting Ross Light (2016-07-13 19:31:18)

>    This looks right to me. I'm at Gophercon right now, and I'd like to
>    write a test to exercise this code path before signing off.

Sounds good. I'll see what I can do re: testing, but wanted to make sure
I roughly understood first.

zombiezen:
Cool. This should suffice for Sandstorm's use, since the promise always resolves to the sender's side, IIUC.  You'll have to write a packet-level test as in rpc_test.go, which isn't as nice as I'd like.  Perhaps those tests can be improved with some of the work I'm doing on the pogs branch.

zenhack:
I parametrized some of the existing tests for different capability types. I'm unsatisifed with the code; having that boolean in there feels wrong, but perhaps you could comment on the general approach?

zenhack:
@zombiezen ping?

rpc/rpc_test.go
```diff
@@ -67,26 +67,78 @@ func readBootstrap(t *testing.T, ctx context.Context, conn *rpc.Conn, p rpc.Tran
...
+// * The message is not of type `Finish`
+// * The message's question ID is not equal to `questionID`.
+//
+// Parameters:
+//
+// t: the *testing.T object for the test in which we're running.
+// ctx: The context to be used when sending the message.
+// p: The rpc.Transport to send the message on.
+// questionID: The expected question ID.
+func recvFinish(t *testing.T, ctx context.Context, p rpc.Transport, questionID uint32) {
```
zombiezen: Instead of taking in a `testing.T`, return an error.

rpc/rpc_test.go
```diff
@@ -67,26 +67,78 @@ func readBootstrap(t *testing.T, ctx context.Context, conn *rpc.Conn, p rpc.Tran
...
+//
+// Parameters:
+//
+// t: the *testing.T object for the test in which we're running.
+// ctx: The context to be used when sending the message.
+// p: The rpc.Transport to send the message on.
+// answerId: The message's answerId.
+// isPromise: If this is true, the capability in the response will be of type
+//   senderPromise, otherwise it will be of type senderHosted.
+func sendBootstrapReturn(t *testing.T, ctx context.Context, p rpc.Transport, answerId uint32, isPromise bool) {
```
zombiezen: Same here. Return an error instead.

zombiezen:
Sorry about the delay, somehow I missed your question.  The tests are a bit grungy because they were the first ones in here I wrote, but you're on the right track.  Just a couple minor comments.

zenhack:
@zombiezen, made the changes. It probably makes sense to propagate it up through bootstrapAndFulfill as well (despite that not being something I did), no? I can add that as well if you like.

I can also squash things, but I tend to find it easier to do review when the history is intact, so I'm holding off.

zenhack:
@zombiezen, ping?

zombiezen:
:+1: Sorry for the delays, these last couple of weeks have been quite busy. I'll merge this in tomorrow (away from dev machine).

zenhack:
Cool. Tangentially: I got hello world + net/http + talking to the sandstorm API working today! and I'm about to open another issue, because hacks.

PR #48, opened 2016-08-25
mem: skip defer for performance

abraithwaite:
We're looking to move to go-capnproto2 from the original and we've been comparing the performance of the two using [alecthomas' go serialization benchmark suite](https://github.com/alecthomas/go_serialization_benchmarks).

This is a minor patch which improves the speed moderately.  We're still looking at options for increasing the speed.  Most of them probably involve changing the go structs to pointers, if that's amenable to you.

Here's the performance comparison with and without this changeset:

https://gist.github.com/abraithwaite/a8876d351edf16bc052cd6b5acdaa330

I'm not sure why the svgs don't render nicely on gist though :-(

### DIFF
--- a/mem.go
+++ b/mem.go
@@ -167,22 +167,27 @@ func (m *Message) NumSegments() int64 {
 
 // Segment returns the segment with the given ID.
 func (m *Message) Segment(id SegmentID) (*Segment, error) {
-	m.mu.Lock()
-	defer m.mu.Unlock()
+	var seg *Segment
 	if isInt32Bit() && id > maxInt32 {
 		return nil, errSegment32Bit
 	}
-	if seg := m.segment(id); seg != nil {
+	m.mu.Lock()
+	if seg = m.segment(id); seg != nil {
+		m.mu.Unlock()
 		return seg, nil
 	}
 	if int64(id) >= m.Arena.NumSegments() {
+		m.mu.Unlock()
 		return nil, errSegmentOutOfBounds
 	}
 	data, err := m.Arena.Data(id)
 	if err != nil {
+		m.mu.Unlock()
 		return nil, err
 	}
-	return m.setSegment(id, data), nil
+	seg = m.setSegment(id, data)
+	m.mu.Unlock()
+	return seg, nil
 }
 
 // segment returns the segment with the given ID.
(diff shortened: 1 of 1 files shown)


zombiezen:
How I wish that defer was faster.  This seems fine.

I remember Jason using that benchmark suite to compare perf numbers before.  You may also be interested in running `go test -bench='BenchmarkUnmarshal_Reuse'`, which makes even less allocations with an increase in potential complexity.  I haven't gotten around to upstreaming those.

I'm curious: why will converting the structs to pointers make perf improvements?  I recently made a largeish API change to remove allocations in the hot paths because it was adversely impacting perf.

abraithwaite:
My mistake.  My expectation is that by not allocating the metadata structs it'll speed it up a bit.

mem.go
```diff
@@ -168,21 +168,27 @@ func (m *Message) NumSegments() int64 {
 // Segment returns the segment with the given ID.
 func (m *Message) Segment(id SegmentID) (*Segment, error) {
 	m.mu.Lock()
-	defer m.mu.Unlock()
+	var seg *Segment
 	if isInt32Bit() && id > maxInt32 {
```
zombiezen: nit: this case doesn't require the mutex

mem.go
```diff
@@ -168,21 +168,27 @@ func (m *Message) NumSegments() int64 {
 // Segment returns the segment with the given ID.
 func (m *Message) Segment(id SegmentID) (*Segment, error) {
 	m.mu.Lock()
```
zombiezen: While we're here, does it make any performance difference to use a RWMutex and have a fast-path where we don't consult the arena and acquire a RLock?

mem.go
abraithwaite: You mean we can move `m.mu.Lock()` below this block correct?

zombiezen: Yup.

mem.go
abraithwaite: This potentially makes sense.  We don't have a use case where we're frequently accessing the data concurrently, but it would certainly make a difference in highly concurrent applications.

To add to that, I wouldn't be surprised if most capnp use cases are write once, read a lot.

abraithwaite:
Coming back to this:

> why will converting the structs to pointers make perf improvements? 

> You may also be interested in running go test -bench='BenchmarkUnmarshal_Reuse'

That's exactly what I was looking for.  Only just got around to reading that code.  I also made a patch for the the serialization benchmarks repo. Here are the results: https://gist.github.com/abraithwaite/905555e81b35eca8d30987690dfc1220

Very nice! 👏

zombiezen:
Merged with some tweaks to critical sections.

You should go ahead and upstream the benchmark changes.

Issue #60, opened 2017-02-17
rpc: allow specifying memory allocation strategy.


zenhack:
Background: the other day @justincormack wrote to the capnproto mailing list about his [memfd][1] package, which includes an implementation of the Arena interface required by go-capnproto2. I took a quick and dirty [whack][2] at implementing rpc.Transport on top of this, so it could actually be used for rpc, but for this to work we'd need to make sure the allocations all happen using memfds, and I don't see a way to get the rpc package to do that.

Unless I'm missing something, it would be nice to add a ConnOption that allows specifying a strategy or such. Maybe just taking a function that returns a new Arena.

Thoughts?

[1]: https://github.com/justincormack/go-memfd
[2]: https://github.com/zenhack/capnp-memrpc-go

justincormack:
There was another thing that came up when doing this which was that `Message.Marshal()` always does append() on the message header and the segments. I could have supported this if `Marshal()` was delegated to the `Arena` as then I could have left space for the transport header at the beginning of the shared memory area, offsetting the segment. As it is I just ignored marshalling as it is not really needed but having it would make it easier to swap out different arena types potentially.

zombiezen:
SGTM. Happy to accept a PR for this.  The approach sounds reasonable: there should be a custom `rpc.Transport` to avoid the call to `Marshal` (can be implemented using existing API) and there should be an `rpc.ConnOption` to specify the arena creation strategy for the `rpc.Conn` (needs a new API).

As for the message header copy, I'm not that familiar with this API, but you could skip the header if you have some other way of communicating segment boundaries.  Glancing at the arena implementation, you're only using one segment, right?

justincormack:
Yes, it works fine without marshalling, there is one segment and the length is available anyway. It would only be useful if it was easier then to write more generic code.

zombiezen:
Addressed in v3

PR #61, opened 2017-02-17
Parametrize rpc alloc

zenhack:
First pass at a solution to #60. Would appreciate comments on both the general design and the stuff in the second commit message.

//cc @zombiezen, @justincormack

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -76,7 +76,7 @@ func (a *answer) fulfill(obj capnp.Ptr) error {
 		}
 		a.queue = nil
 	} else {
-		retmsg := newReturnMessage(nil, a.id)
+		retmsg := newReturnMessage(a.id)
 		ret, _ := retmsg.Return()
 		payload, _ := ret.NewResults()
 		payload.SetContentPtr(obj)
@@ -116,7 +116,7 @@ func (a *answer) reject(err error) error {
 		panic("answer.reject called more than once")
 	}
 	a.err, a.done = err, true
-	m := newReturnMessage(nil, a.id)
+	m := newReturnMessage(a.id)
 	mret, _ := m.Return()
 	setReturnException(mret, err)
 	var firstErr error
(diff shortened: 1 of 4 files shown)


zombiezen:
This changes three assumptions that the rpc package has made:

1. Creating a message will not fail.
2. Writing to an arena inside the RPC package will not fail.
2. Arenas don't need any other finalization other than GC.

The `capnp` package doesn't make these assumptions, which is good.  I'm not sure how much complexity this will add to the rpc package if it no longer makes these assumptions.  If it really complicates things, I'm not sure how I want to proceed.  But let's keep going for now until we see.

I'd imagine that the allocator should actually be more like this:

```go
// An Allocator creates arenas for sending messages.
//
// Allocate returns an arena that is distinct from any of the other in-use arenas
// or an error if one can't be provided (e.g. memory limits).  The `done` function
// will be called once the caller is no longer using any of the memory in the arena.
type Allocator interface {
  Allocate() (arena capnp.Arena, done func(), err error)
}
```

zenhack:
(1) and (2) don't seem like they would cause much of a problem to handle; we can just drop the connection and return an error when those happen (and call done where needed).

I'm noticing I missed a couple places where the rpc package explicitly calls SingleSegment or MultiSegment, which obviously needs to change.

re: (3), for received messages the transport allocates the memory (and thus needs to have it's own reference to an allocator) and continues to own it, so it can be responsible for calling done(). The obvious way for an implementation to handle this (if needed) is to just call done() on the previous message at the top of RecvMessage, and then on the last message in Close(). The interface doesn't specify who owns the message after the call to SendMessage returns, but if we specify that the caller owns it, we may be able to just call done() whenn SendMessage() returns, though I don't know if this will affect transitive callers.

Is there anywhere that the rpc package allocates but doesn't pass the result through a transport?

zenhack:
So, noticed that the implementation makes full copies of call messages. This seems like it would make the in-memory rpc basically moot, so perhaps we should step back and think about whether some variant of this solution is even adequate for the end goal, or if not what other changes we need for this to be useful.

zenhack:
Observation: for @justincormack's approach to the rpc, the transport contract is sortof the opposite of what we really want -- for that transport it would make more sense if the caller of RecvMessage owned the return value, and calling SendMessage transfers ownership to the transport.

zombiezen:
I've been thinking about this for a couple months, and I think what you would have to address is that received messages legitimately need to be owned by application code past `server.Ack`.  Having to do a copy for these cases seems onerous.

zenhack:
Yeah, that's kindof what I was getting at with my above comment. If we could just invert the contract as I described above, we could avoid doing the copy on receive, and close the fds on send. Sends and receives should be reasonably balanced, so this should make sure that as long as we limit the number of active messages, we should be able to avoid over-using fds. It doesn't seem to me like it would be too onerous to specify that you can't hold on to stuff you shove into a `*_Result`. There's of course a question of breaking compatibility.

zombiezen:
I appreciate the design discussion here and I'm going to think about this as I work on v3.  I'm going to close out this PR because I think that it's going to look differently after the changes.

zenhack:
Fair enough.

Issue #62, opened 2017-02-19
pogs: support interface types


zenhack:
I'm trying to write some table-driven tests for go.sandstorm, and the pogs package makes this a lot easier, but it's still a bit annoying because I can't put capnp interface types in the struct literals. I don't think this should be hard to implement; we can require that capnp interface fields to have an rpc.Client field, and it should just be an assignment.

zombiezen:
Yup, this sounds right. I omitted it just for sake of time, not really as a "it shouldn't be done".

zenhack:
It occurs to me that implementing this properly requires access to the capability table, which *is* transitively accessable via capnp.Struct (segment -> message), but Struct.seg isn't exported, so right now this doesn't quite work from an purely external package.

(I know you're busy this week, and don't really expect a timely response, but I figure there's no harm in frontloading).

zombiezen:
All structs have a `Segment()` method. You can always get the segment/message, you just can't modify the `seg` field.

zenhack:
Ah, missed that. should do the trick then.

zenhack:
(Just a note: I've got a branch where I'm working on this).

zenhack:
#66 is merged, now, so I'm closing this.

PR #66, opened 2017-02-24
Pogs interfaces

zenhack:
Addresses #62.

//cc @zombiezen, when you have the time.


pogs/extract.go
```diff
@@ -25,6 +25,14 @@ type extracter struct {
 	nodes nodemap.Map
 }
 
+var (
```
zombiezen: Don't really need the doc comment.

```go
var clientType = reflect.TypeOf((*capnp.Client)(nil)).Elem()
```

pogs/extract.go
```diff
@@ -219,6 +227,19 @@ func (e *extracter) extractField(val reflect.Value, s capnp.Struct, f schema.Fie
...
 		}
 		return e.extractList(val, typ, l)
+	case schema.Type_Which_interface:
+		p, err := s.Ptr(uint16(f.Slot().Offset()))
+		if err != nil {
+			return err
+		}
+		client := p.Interface().Client()
+		if client != nil {
+			if !val.Type().Implements(clientType) {
```
zombiezen: I think you actually want

```go
if val.Type() != clientType {
```

pogs/extract.go
```diff
@@ -376,6 +397,15 @@ func isTypeMatch(r reflect.Type, s schema.Type) bool {
 	case schema.Type_Which_list:
 		e, _ := s.List().ElementType()
 		return r.Kind() == reflect.Slice && isTypeMatch(r.Elem(), e)
+	case schema.Type_Which_interface:
+		if !r.Implements(clientType) && r.Kind() == reflect.Struct {
+			field, ok := r.FieldByName("Client")
```
zombiezen: I'd like for this to check for a more strict bound:

1. Has one field
2. That field is called `Client`
3. That field is of type `capnp.Client`

pogs/insert.go
```diff
@@ -218,12 +218,32 @@ func (ins *inserter) insertField(s capnp.Struct, f schema.Field, val reflect.Val
...
+		if err := s.SetPtr(off, ptr); err != nil {
+			return err
+		}
 	default:
 		return fmt.Errorf("unknown field type %v", typ.Which())
 	}
 	return nil
 }
 
+func getCapPtr(seg *capnp.Segment, val reflect.Value) capnp.Ptr {
```
zombiezen: nit: just `capPtr`

pogs/interface_test.go
```diff
@@ -0,0 +1,168 @@
+package pogs
+
+import (
+	"golang.org/x/net/context"
+	"testing"
```
zombiezen: nit: group stdlib packages apart from others

pogs/interface_test.go
```diff
@@ -0,0 +1,168 @@
...
+
+type EchoBases struct {
+	Bases []EchoBase
+}
+
+type Hoth struct {
+	Base EchoBase
+}
+
+func TestInsertIFace(t *testing.T) {
```
zombiezen: Any reason to not add these to the broader table test?  I think I have better error handling in there.

zenhack: Using reflect.DeepEqual on interfaces doesn't really do what we want in this case, and it seemed prudent to have something that tested the behavior of the interface, which would require special casing in the general test anyway. I think at the time I decided making it work would be more disruptive to the existing tests than really made sense. I'm willing to be convinced otherwise.

zenhack: @zombiezen, I think I've addressed everything besides this point. Any thoughts on my comment above?

zenhack: @zombiezen, ping?

zombiezen:
Sorry, lost track of this with other stuff. I'll take a look tomorrow.

pogs/extract.go
```diff
@@ -376,6 +391,24 @@ func isTypeMatch(r reflect.Type, s schema.Type) bool {
 	case schema.Type_Which_list:
 		e, _ := s.List().ElementType()
 		return r.Kind() == reflect.Slice && isTypeMatch(r.Elem(), e)
+	case schema.Type_Which_interface:
+		if r.Implements(clientType) {
```
zombiezen: I think you want `r == clientType` here.

zenhack:
I think I can make the tests work with the table; I'll have to add a bit of logic to actually call testEcho when the interface is (expectedly) not nil. So it won't be checking for equality really, but that's OK I think.

I addressed your other two comments.

zenhack:
@zombiezen, I added interface stuff to the main table based tests. It's actually failing currentlly in a case where I'm trying to insert a slice of structs, where some of them have a nil client and some don't; the resulting message ends up with them all filled in.

I bet I know what's going on: the composite list wire format can't actually represent lists of structs that have different numbers of pointers filled in, so there's an extra pointer at the end of some of the structs that's being interpreted as a capability pointer. CapNProto doesn't have a notion of a nil/null pointer, just an "absent" one, so I don't think there's a "correct" way to encode a list like this.

I'm not sure how to handle this case besides just mentioning in the docs that you can't do that. Thoughts?

zombiezen:
> I bet I know what's going on: the composite list wire format can't actually represent lists of structs that have different numbers of pointers filled in, so there's an extra pointer at the end of some of the structs that's being interpreted as a capability pointer.

But there is a null pointer for capabilities, this happens all over the place.  It's the zero word.  That seems like a legitimate failure to me.  Any pointer type in Cap'n Proto can be null.

zenhack:
Ah, you're right -- I missed that reading the spec. I will dig a bit deeper then.

zenhack:
@zombiezen, fixed.

zombiezen:
Thanks! Sorry this took a while, but I'm happy with how it turned out.

zenhack:
Cool. No worries about the delays; things come up.

Issue #68, opened 2017-03-07
Use net.Buffers in Go 1.8 for encoding.


tmthrgd:
Go 1.8 [added](https://golang.org/doc/go1.8#net) a new [`net.Buffers`](https://golang.org/pkg/net/#Buffers) designed specifically for writing multiple blocks of memory to a network connection in a single system call.

> The new Buffers type permits writing to the network more efficiently from multiple discontiguous buffers in memory. On certain machines, for certain types of connections, this is optimized into an OS-specific batch write operation (such as writev).

As `net.Buffers` is essentially a 'best effort' system, writes to regular `io.Writer`'s will function identically without any change in functionality or semantics.

`writev` also ensures that the write performed will be atomic, although this **cannot** be relied upon by `net.Buffers` as it doesn't guarantee the use of `writev`. According to the [man page](https://linux.die.net/man/2/writev):

> The data transfers performed by readv() and writev() are atomic: the data written by writev() is written as a single block that is not intermingled with output from writes in other processes (but see pipe(7) for an exception).

I think the principal benefit of using `net.Buffers` would be reducing the number of system calls required in [(*Encoder).Encode](https://github.com/zombiezen/go-capnproto2/blob/02f10324cc25fc4acd262b854872e1e5bddb14f7/mem.go#L567-L599) from `1 + number of segments` to 1 in every case – with the exception of the packed encoder as it writes the packed data to a [single temporary buffer](https://github.com/zombiezen/go-capnproto2/blob/02f10324cc25fc4acd262b854872e1e5bddb14f7/mem.go#L602-L605).

zombiezen:
Sounds like a good idea. Happy to accept PRs.

PR #70, opened 2017-03-17
Use net.Buffers in Go 1.8 for encoding.

tmthrgd:
This pull request covers issue #68.

On go1.7 and below it behaves as before. On go1.8 with the packed encoder it also behaves as before. On go1.8 with the regular encoder it uses [`net.Buffers`](https://golang.org/pkg/net/#Buffers) which will use [`writev`](https://linux.die.net/man/2/writev) in a single system call where possible.

It adds go1.8 to travis to test the new code path.


mem.go
```diff
@@ -583,28 +583,16 @@ func (e *Encoder) Encode(m *Message) error {
 		e.hdrbuf = e.hdrbuf[:hdrSize]
 	}
 	marshalStreamHeader(e.hdrbuf, sizes)
-	if err := e.write(e.hdrbuf); err != nil {
-		return err
-	}
+	bufs := make([][]byte, 1+nsegs)
```
zombiezen: Seems like this list of bufs could be reused between calls.

mem_18.go
```diff
@@ -0,0 +1,23 @@
...
+
+package capnp
+
+import (
+	"net"
+
+	"zombiezen.com/go/capnproto2/internal/packed"
+)
+
+func (e *Encoder) write(bufs net.Buffers) error {
```
zombiezen: Keep the function parameter type the same between 1.8 and non-1.8.

mem_other.go
```diff
@@ -0,0 +1,18 @@
+// +build !go1.8
+
+package capnp
+
+import "zombiezen.com/go/capnproto2/internal/packed"
+
+func (e *Encoder) write(bufs [][]byte) error {
+	for _, b := range bufs {
+		if e.packed {
```
zombiezen: Pull out the `e.packed` logic into a separate (non-tagged) function.

mem.go
zombiezen: GH is not letting me add a comment farther up... while you're visiting this code, it seems like traversing the segments twice (using `segmentSizes`) isn't really necessary.  Mind rolling that all into one?

mem_other.go
tmthrgd: Done.

mem_18.go
tmthrgd: Done.

mem.go
tmthrgd: Done.

tmthrgd: Done.

mem.go
```diff
@@ -571,40 +573,51 @@ func (e *Encoder) Encode(m *Message) error {
...
 	}
-	return nil
+	maxSeg := uint32(nsegs - 1)
+	hdrSize := streamHeaderSize(maxSeg)
+	if uint64(cap(e.hdrbuf)) < hdrSize {
+		e.hdrbuf = make([]byte, hdrSize)
+	} else {
+		e.hdrbuf = e.hdrbuf[:hdrSize]
+	}
+	marshalStreamHeader(e.hdrbuf, e.sizes)
```
zombiezen: TBH, I think I want to nuke `marshalStreamHeader`.  It's adding an allocation for not much clarity.  Let's inline it here (and define a new helper, `appendUint32`):

```go
if uint64(cap(e.hdrbuf)) < hdrSize {
  e.hdrbuf = make([]byte, 0, hdrSize)
}
e.hdrbuf = appendUint32(e.hdrbuf[:0], maxSeg)
for i, buf := range e.bufs {
 e.hdrbuf = appendUint32(e.hdrbuf, uint32(Size(len(buf))/wordSize))
}
if len(e.hdrbuf) % int(wordSize) != 0 {
  e.hdrbuf = appendUint32(e.hdrbuf, 0)
}
```

Interestingly, I think that the current `Encoder` has a padding bug in this regard.  Can you add a test where a 3-segment message is encoded then a 2-segment message is encoded, and then assert that the 2-segment message's header properly contains zero-padding (instead of garbage)?

mem_18.go
```diff
@@ -0,0 +1,10 @@
+// +build go1.8
+
+package capnp
+
+import "net"
+
+func (e *Encoder) write(bufs [][]byte) error {
+	_, err := (*net.Buffers)(&bufs).WriteTo(e.w)
```
zombiezen: No need for pointer indirection, just convert to `net.Buffers`.

mem.go
```diff
@@ -571,40 +573,51 @@ func (e *Encoder) Encode(m *Message) error {
...
+		e.sizes = e.sizes[:nsegs]
 	}
 	for i := int64(0); i < nsegs; i++ {
 		s, err := m.Segment(SegmentID(i))
 		if err != nil {
 			return err
 		}
-		if err := e.write(s.data); err != nil {
-			return err
+		n := len(s.data)
```
tmthrgd: That would risk an overflow that the `int64(n) > int64(maxSize)` is designed to avoid. Size is a `uin32` but `len(x)` returns an `int` which can easily be larger than a `uint32` can handle.

mem_18.go
tmthrgd: I tried that before but (*net.Buffers).WriteTo is a pointer reciver.

mem.go
```diff
@@ -571,40 +573,51 @@ func (e *Encoder) Encode(m *Message) error {
 	if nsegs == 0 {
```
tmthrgd: Done.

mem.go
```diff
@@ -571,40 +573,51 @@ func (e *Encoder) Encode(m *Message) error {
...
-	sizes, err := m.segmentSizes()
-	if err != nil {
-		return err
-	}
-	maxSeg := uint32(nsegs - 1)
-	hdrSize := streamHeaderSize(maxSeg)
-	if uint64(cap(e.hdrbuf)) < hdrSize {
-		e.hdrbuf = make([]byte, hdrSize)
+	if int64(cap(e.bufs)) < 1+nsegs {
+		e.bufs = make([][]byte, 1+nsegs)
```
tmthrgd: Done.

mem.go
```diff
@@ -571,40 +573,51 @@ func (e *Encoder) Encode(m *Message) error {
...
 			return err
 		}
-		if err := e.write(s.data); err != nil {
-			return err
+		n := len(s.data)
+		if int64(n) > int64(maxSize) {
+			return errSegmentTooLarge
 		}
+		e.sizes[i] = Size(n)
+		e.bufs[1+i] = s.data
```
tmthrgd: Done.

mem.go
tmthrgd: I've been able to reproduce that bug and have added a failing test case for it. I'll inline marshalStreamHeader and fix the bug now.

tmthrgd: I've only touched the `(*Encoder).Encode` part so far and I've left the use of `marshalStreamHeader` in `(*Message).Marshal`. So the bug still exists in `(*Message).Marshal`.

I'm not sure how you want me to handle that, but I'm thinking it would be best if `(*Message).Marshal` and `(*Message).MarshalPacked` were just made wrappers around `(*Encoder).Encode`. That would mean the encoding logic only exists once and would reduce duplication. I think I'll go ahead and add a commit for that and you can decide whether you like that method or not.

mem_18.go
zombiezen: How bizarre. Opened golang/go#19680

mem.go
zombiezen: How does the bug exist in Marshal?  The issue is that the buffer gets reused and there's no buffer to reuse for Marshal.

IIRC, having different codepaths does have a big performance impact (i.e. Marshal is faster when it doesn't go through a Writer).  I would rather keep the separation unless you can prove that this doesn't significantly alter performance.

I realize this is a bit bigger of a change than you probably wanted to do.  Let me know if you want me to pick up this part.

mem_test.go
```diff
@@ -708,6 +708,47 @@ func TestDecoder_MaxMessageSize(t *testing.T) {
 	}
 }
 
+func TestStreamHeaderPadding(t *testing.T) {
```
zombiezen: Add a comment noting that this is a regression check for header buffer padding.

tmthrgd: Done.

mem_18.go
tmthrgd: It's a result of how `net.Buffers` is implemented so it's very intentional. `net.Buffers` needs to work with multiple calls to `Read` and `WriteTo` (for things like `io.Copy`), to do this it needs to be able to modify the slice as it consumes more data - see [src/net/net.go](https://github.com/golang/go/blob/25a1d5d0f472f0da82aa8b4a1f3322463df5bbe6/src/net/net.go#L677).

mem.go
tmthrgd: I backed that change out.

I think I misread you're desire to 'nuke' `marshalStreamHeader` as a request *for me* to eliminate it. I'll leave removing `marshalStreamHeader` up to you as a separate matter.

PR #71, opened 2017-03-19
Far far pointer bug fix

levrado:
This pull request covers issue #72.

### DIFF
--- a/rawpointer.go
+++ b/rawpointer.go
@@ -56,10 +56,8 @@ func rawDoubleFarPointer(segID SegmentID, off Address) rawPointer {
 // a near pointer in the destination segment.  Its offset will be
 // relative to the beginning of the segment.
 func landingPadNearPointer(far, tag rawPointer) rawPointer {
-	farFarAddress := far.farAddress() / Address(wordSize)
-
 	// remove 1 word and we'll get the offset to the start of the data section
-	rawFarFarPointer := rawPointer(farFarAddress - 1)
+	rawFarFarPointer := rawPointer(far.farAddress() / Address(wordSize) - 1)
 	
 	// finally to get an actual pointer do an OR with the tag after moving rawFarFarPointer
 	// by 2 bits which are the type indicator bits
--- a/rawpointer.go
+++ b/rawpointer.go
@@ -58,12 +58,11 @@ func rawDoubleFarPointer(segID SegmentID, off Address) rawPointer {
 func landingPadNearPointer(far, tag rawPointer) rawPointer {
 	farFarAddrress := far.farAddress() / Address(wordSize)
 
-	// remove 1 word and we'll get the offset from the end of the pointer to the
-	// start of the struct's data section
+	// remove 1 word and we'll get the offset to the start of the data section
 	rawFarFarPointer := rawPointer(farFarAddrress - 1)
-
-	// move 2 bits to make rawFarFarPointer like a struct pointer (2 lsb's are zero),
-	// finally to get a struct pointer do an OR with the tag
+	
+	// finally to get an actual pointer do an OR with the tag after moving rawFarFarPointer
+	// by 2 bits which are the type indicator bits
 	return tag | rawFarFarPointer << 2
 }
 
--- a/rawpointer.go
+++ b/rawpointer.go
@@ -56,7 +56,15 @@ func rawDoubleFarPointer(segID SegmentID, off Address) rawPointer {
 // a near pointer in the destination segment.  Its offset will be
 // relative to the beginning of the segment.
 func landingPadNearPointer(far, tag rawPointer) rawPointer {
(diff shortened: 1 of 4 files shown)


rawpointer.go
```diff
@@ -56,7 +56,14 @@ func rawDoubleFarPointer(segID SegmentID, off Address) rawPointer {
 // a near pointer in the destination segment.  Its offset will be
 // relative to the beginning of the segment.
 func landingPadNearPointer(far, tag rawPointer) rawPointer {
-	return tag | rawPointer(far.farAddress()-Address(wordSize))<<2
+	farFarAddrress := far.farAddress() / Address(wordSize)
```
zombiezen: typo: farFarAddress

rawpointer_test.go
```diff
@@ -232,3 +232,39 @@ func TestRawPointerTotalListSize(t *testing.T) {
 		}
 	}
 }
+
+func TestLandingPadNearPointer(t *testing.T) {
```
zombiezen: This test doesn't really help me understand what failed with the previous implementation.  Instead, can you write this as a table test of:

```go
tests := []struct{
  far     rawPointer
  tag     rawPointer
  landing rawPointer
}{
  // ...
}
```

The test as written is halfway between a unit test and an integration test (usually a bad thing): it checks results by using other functions instead of testing just this function.

It would be nice to have a small integration test as well.  I know far pointers don't have as great test coverage, so it would help to have more.

levrado:
fixed according to comments, thanks for reviewing!

rawpointer_test.go
```diff
@@ -232,3 +232,24 @@ func TestRawPointerTotalListSize(t *testing.T) {
...
+		{0xa0, 0x2000200000000, 0x200020000004c},
+		{0xb12, 0x2000200000000, 0x2000200000584}, // struct pointer
+		{0xb12, 0x2000200000001, 0x2000200000585}, // list pointer
+		{0xb12, 0x2000200000003, 0x2000200000587}, // capability pointer
+	}
+
+	for _, test := range tests {
+		testNearPointer := landingPadNearPointer(test.far, test.tag)
+		if testNearPointer != test.landing {
+			t.Errorf("rawPointer(%#016x).pointerType() = %d; want rawPointer(%#016x).pointerType() = %d", testNearPointer, testNearPointer.pointerType(), test.landing, test.landing.pointerType())
```
zombiezen: I've already spent a fair amount of work making rawPointers values easy to debug.  Please rewrite this as:

```go
t.Errorf("landingPadNearPointer(%v, %v) = %v; want %v", test.far, test.tag, landing, test.landing)
```

rawpointer_test.go
```diff
@@ -232,3 +232,24 @@ func TestRawPointerTotalListSize(t *testing.T) {
...
 	}
 }
+
+func TestLandingPadNearPointer(t *testing.T) {
+	tests := []struct {
+		far     rawPointer
+		tag     rawPointer
+		landing rawPointer
+	}{
+		{0x08, 0x2000200000000, 0x2000200000000},
```
zombiezen: These tests would also benefit in being rewritten in terms of rawPointer constructors.  For example, this first one would be:

```go
{rawFarPointer(0, 8), rawStructPointer(0, ObjectSize{16, 2}), rawStructPointer(0, ObjectSize{16, 2})},
```

rawpointer_test.go
zombiezen: This is much better, thank you!

rawpointer.go
```diff
@@ -56,7 +56,14 @@ func rawDoubleFarPointer(segID SegmentID, off Address) rawPointer {
 // a near pointer in the destination segment.  Its offset will be
 // relative to the beginning of the segment.
 func landingPadNearPointer(far, tag rawPointer) rawPointer {
-	return tag | rawPointer(far.farAddress()-Address(wordSize))<<2
+	farFarAddress := far.farAddress() / Address(wordSize)
```
zombiezen: Since this variable is not a byte address (it's not a word address), let's inline this into the next line.

rawpointer_test.go
levrado: Hey, trying to do as you suggest here but I have a problem understanding something, would love if you could clarify, in the function `rawFarPointer` you do:
`return farPointer | rawPointer(off&^7) | (rawPointer(segID) << 32)`. 
why the AND NOT in here? aren't you going against the spec here?

for example if I pass an offset of 160 to the function what I'll get is 20 in the offset section according to capnp spec and not the expected 160.

shouldn't be `return farPointer | rawPointer(off) << 3 | (rawPointer(segID) << 32)`?

zombiezen: I've defined `Address` to be a byte address, not a word address.  `rawFarPointer` is doing the conversion and bit-shifting at the same time by just lopping off the bottom three bits.

levrado:
updated :)

PR #73, opened 2017-03-26
Add 'Code generated by capnp.' comment to capnpc-go

tmthrgd:
GitHub currently renders the full diff of the generated .capnp.go which is messy. When .go files have a `Code generated by` comment as the first line, the diffs will be suppressed by default.

This pull request adds such a comment to avoid polluting the diffs with automatically generated changes. It does nothing else.

(I haven't added myself to AUTHORS+CONTRIBUTORS because I did that in #70 which is still to be merged).


zombiezen:
Nice, thanks!

Issue #79, opened 2017-05-14
Reduce allocations in streaming decoder


tj:
First of all this is awesome! Thanks a lot for the package.

I haven't looked heavily yet but the [msgp](https://github.com/tinylib/msgp) package seems to decode roughly ~70% faster given the same structure, which seems a little odd since one goal of capnproto is to reduce the allocations (and msgp has to construct everything).

I was thinking maybe splitting the framing out of the decoder so that Unmarshal can be used, allowing the user to decide whether or not it's safe to reference the memory. My main use-case for analytical type stuff where you just want to iterate and grab a field or two in-place.

Any thoughts? Maybe I missed an API in there, but it looked like the framing was built directly into the decoder.

IIRC msgp allows you to unmarshal against an existing byte slice (or allocs if necessary), I think it makes use of that pretty heavily internally which might be why it's quicker in this case. 

cheers

zombiezen:
Do you have a Go-style benchmark where you can show the issue?  This library doesn't have nearly enough.

PR #82, opened 2017-05-16
Add benchmarks

tj:
RE #79 

Hopefully not misusing the API in the bench. Output on my MBP illustrating the comparison with msgp, even without decoding any fields from the capn msg. 

```
BenchmarkUnmarshal-8             	 2000000	       777 ns/op	     208 B/op	       3 allocs/op
BenchmarkUnmarshal_Reuse-8       	 5000000	       282 ns/op	       0 B/op	       0 allocs/op
BenchmarkUnmarshalMsgp-8         	 5000000	       281 ns/op	      32 B/op	       2 allocs/op
BenchmarkUnmarshalMsgpReader-8   	     500	   3299178 ns/op	 294.01 MB/s	  322376 B/op	   20004 allocs/op
BenchmarkUnmarshalDecoder-8      	     300	   5520729 ns/op	 173.89 MB/s	 3200052 B/op	   60002 allocs/op
```

Hopefully I'll have some time this week to check out a solution if you don't get to it first! Looks like it should be comparable if all else is equal.

### DIFF
--- a/msgp_bench_test.go
+++ b/msgp_bench_test.go
@@ -1,4 +1,6 @@
 // +build msgp
+//go:generate msgp -tests=false -o msgp_bench_gen_test.go
+//msgp:Tuple Event
 
 package capnp_test
 
@@ -14,9 +16,6 @@ import (
 
 	"github.com/tinylib/msgp/msgp"
 )
-
-//go:generate msgp -tests=false -o msgp_bench_gen_test.go
-//msgp:Tuple Event
 
 type Event struct {
 	Name     string
--- a/msgp_bench_test.go
+++ b/msgp_bench_test.go
@@ -8,6 +8,9 @@ import (
 	"testing"
 	"time"
 
+	capnp "zombiezen.com/go/capnproto2"
+	air "zombiezen.com/go/capnproto2/internal/aircraftlib"
+
 	"github.com/tinylib/msgp/msgp"
 )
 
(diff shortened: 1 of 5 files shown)


msgp_bench_test.go
```diff
@@ -0,0 +1,127 @@
...
+type Event struct {
+	Name     string
+	BirthDay time.Time
+	Phone    string
+	Siblings int
+	Spouse   bool
+	Money    float64
+}
+
+func BenchmarkUnmarshalMsgp(b *testing.B) {
```
zombiezen: Mind putting the MsgPack-related files and benchmarks behind a build tag?  I don't want the tests to depend on MsgPack unless specifically requested.

msgp_bench_test.go
```diff
@@ -0,0 +1,127 @@
...
+		for {
+			var e Event
+			err := e.DecodeMsg(r)
+
+			if err == io.EOF {
+				break
+			}
+
+			if err != nil {
+				log.Fatal(err)
```
zombiezen: Use `b.Fatal` instead (here and elsewhere).

msgp_bench_test.go
```diff
@@ -0,0 +1,127 @@
...
+			}
+		}
+	}
+}
+
+func BenchmarkUnmarshalDecoder(b *testing.B) {
```
zombiezen: I'd rename this to BenchmarkDecode.

msgp_bench_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	for i := 0; i < b.N; i++ {
+		dec := capnp.NewDecoder(bytes.NewReader(blob))
+
+		for {
+			msg, err := dec.Decode()
+
+			if err == io.EOF {
+				break
+			}
```
zombiezen: I'd also add an error check here for non-`io.EOF` errors.

zombiezen:
Yeah, those numbers are not what I'd expect at all.  Good find!

tj:
think I got everything, LMK if you want the capn benchmarks in a different file to just isolate the msgp ones

zombiezen:
For now, let's just use `+build msgpbench`.

msgp_bench_test.go
```diff
@@ -0,0 +1,131 @@
...
+
+			if err != nil {
+				b.Fatal(err)
+			}
+		}
+	}
+}
+
+func BenchmarkDecode(b *testing.B) {
```
zombiezen: Let's pull out BenchmarkDecode into integration_test.go.

PR #118, opened 2018-06-15
Update Bazel WORKSPACE and build files

jayconrod:
This change makes this repository compatible with newer versions of rules_go and Bazel. The big differences are that`importpath` attribute is now mandatory in `go_library` rules, and `go_prefix` has been removed.

This change is broken up into two commits to simplify review:

* Add rules_go 0.12.1 and Gazelle 0.12.0 (latest versions) to WORKSPACE. Remove deprecated calls. Add `//:gazelle` rule.
* Run `gazelle fix`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: go
 go_import_path: zombiezen.com/go/capnproto2
 go:
-- 1.4.x
 - 1.x
 install: _travis/install.bash
 script: _travis/build.bash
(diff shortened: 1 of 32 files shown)


jayconrod:
@zombiezen Hopefully this isn't too invasive. Let me know if I should fix anything. This was in response to bazelbuild/rules_go#1551. I figured updating this repo was the best thing to do.

zombiezen:
Ah, I forgot that Travis tests against ancient Go versions. You can remove line 4 from .travis.yml, and I think that should make tests pass.

jayconrod:
Took me a few tries to get Travis CI working. Bazel will now use whatever version of Go is installed in the Travis CI container instead of downloading the equivalent version. I think that's what's intended, but let me know if it should switch back to the download.

PR #125, opened 2018-10-14
fix grammar

DuckerMan:


README.md
```diff
@@ -57,7 +57,7 @@ See the docs on [godoc.org][godoc].
 
 ## What is Cap'n Proto?
 
-The best cerealization...
+The best serealization...
```
zenhack: This bit is intentional; you'll note it's spelled the same way at the top of the Cap'n Proto website:

https://capnproto.org/

Along with the logo, it's a play on Cap'n Crunch:

https://en.wikipedia.org/wiki/Cap'n_Crunch

zombiezen:
Thanks for the PR @DuckerMan, but both of these sentences are correct as-is. Closing.

PR #131, opened 2019-01-11
Allow a message to be marked as "safe" to prevent read limiter tracking

nikhil93uf:
There are cases (both as a feature and as a bug-fix) where read limiter tracking
needs to be disabled to prevent hitting traversal limits. Particularly, auto-generated
vars not-only hit traversal limits very quickly (if referenced extensively) but
also don't need this check since the attack this check is supposed to protect
against (amplification attacks), is not a possibilty here.


zenhack:
Instead of adding a separate `MarkSafe`, you could just do `msg.Reset((1 << 64) - 1)`. In practice you'll never hit that limit. I believe the C++ implementation doesn't provide another way to do this either, and that's been the standard advice.

nikhil93uf:
Agreed. But if the intent is to not check this limit for messages known to be safe from the attack, I'd think we'd want to avoid the overhead of even tracking the read counter. Then Read Limiter.canRead() can be a no-op and just return true instead of doing a bunch of math. And in case of auto-generated vars that are probably used all over the code, this can improve performance

zombiezen:
I agree with @zenhack here: doing `msg.Reset((1<<64) - 1)` is the right way to go. Reducing the number of configurations to test is ideal for this code.

araxus:
When used in a networking dataplane where constant structures to recognize common objects or tags, are defined by the Cap'n grammar, and auto-generated by the compiler ... any use of ReadLimiter.canRead() for data which does not originate from the network is an observable performance hit since they're asserting the LOCK mnemonic with CMPXCHG.  It doesn't make any sense since the feature was introduced as a security mitigation against an amplification attack.

araxus:
No; even at math.MaxUInt64, you're locking and invalidating the CPU cache line on every constant read/comparison.  Since this code is a security mitigation and we know the data is safe, it seems logical to have a way to disable the check entirely for these Messages.

The alternative is to compare against the raw byte slice produced by the compiler, which seems like an ugly hack.

nikhil93uf:
I concur. And I was asking @zombiezen if that was the alternative he was suggesting. The perf impact is obvious here and unnecessary. Also, marking the message as safe is explicit and only enabled by default  for Capnp constants. I don't think this is going to make testing complex.

zombiezen:
@nikhil93uf Ah, I missed the fact that this was on Cap'n Proto constants. Sorry. I am okay to move forward with `.capnp.go` files generating `.Segment().Message().Reset((1<<64) - 1)`.

I understand the perf impact, but I'm weighing it with the complexity of testing bugs and users incorrectly marking messages as safe. Please open a separate issue for that aspect.

nikhil93uf:
Looks like there was an issue with the CI. Is there a way to restart the test?

nikhil93uf:
Closing and re-opening the PR to trigger a re-build

nikhil93uf:
Nvm. The build w/ Bazel seems to broken for months now. @zombiezen Can you please review this PR?

zombiezen:
Yeah, I probably need to rip out the Bazel build. I'll try to take a look this weekend.

capnpc-go/templates/constants
```diff
@@ -12,3 +12,11 @@ var (
 {{end}}
 )
 {{end}}
+
+{{with .Vars -}}
+func init() {
+   // Set traversal limit for constants as Uint64Max since they're safe from amplification attacks
+  {{range .}} {{.Name}}.Segment().Message().ReadLimiter().Reset((1<<64) - 1)
```
zombiezen: I'd like to avoid the generated blank line. Try this:

```
	// Set traversal limit for constants as Uint64Max since they're safe from amplification attacks.
	{{- range .}}
	{{.Name}}.Segment().Message().ReadLimiter().Reset((1<<64) - 1)
	{{- end}}
}
```

PR #140, opened 2019-05-11
standardize .capnp syntax in doc.go

arnottcr:
.capnp examples either used two spaces ("  ") or tabs ("	") for
whitespace and one is incorrectly indented breaking the godoc
formatting. The latter was fixed, and all examples were moved to two
space indenting.

go.mod was formatted, tidied, and a default go directive was added.


zombiezen:
Thanks!

PR #143, opened 2019-07-16
Add option to indent output in encoding/text

martin-sucha:
Indented output is more readable.

### DIFF
--- a/encoding/text/marshal.go
+++ b/encoding/text/marshal.go
@@ -42,20 +42,26 @@ func MarshalList(typeID uint64, l capnp.List) (string, error) {
 
 // An Encoder writes the text format of Cap'n Proto messages to an output stream.
 type Encoder struct {
-	w     errWriter
+	w     indentWriter
 	tmp   []byte
 	nodes nodemap.Map
 }
 
 // NewEncoder returns a new encoder that writes to w.
 func NewEncoder(w io.Writer) *Encoder {
-	return &Encoder{w: errWriter{w: w}}
+	return &Encoder{w: indentWriter{w: w}}
 }
 
 // UseRegistry changes the registry that the encoder consults for
 // schemas from the default registry.
 func (enc *Encoder) UseRegistry(reg *schemas.Registry) {
 	enc.nodes.UseRegistry(reg)
+}
+
+// SetIndent sets string to indent each level with.
+// An empty string disables indentation.
+func (enc *Encoder) SetIndent(indent string) {
+	enc.w.indentPerLevel = indent
 }
 
 // Encode writes the text representation of s to the stream.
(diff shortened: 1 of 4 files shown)


encoding/text/marshal.go
```diff
@@ -453,3 +544,65 @@ func (ew *errWriter) WriteByte(b byte) error {
 	}
 	return ew.err
 }
+
+// indentWriter is helper for writing indented text
+type indentWriter struct {
+	w errWriter
```
zombiezen: Rather than compose the `indentWriter` from an `errWriter`, let's inline the `errWriter`'s functionality into this struct.

encoding/text/marshal_test.go
```diff
@@ -20,25 +20,34 @@ func TestEncode(t *testing.T) {
 	tests := []struct {
 		constID uint64
 		text    string
+		indent  int
```
zombiezen: I'd like to increase the coverage of this new indent mode. Rather than have a separate parameter, let's have each test case include both the indented and not-indented versions.

encoding/text/marshal.go
```diff
@@ -297,98 +312,174 @@ func codeOrderFields(s schema.Node_structNode) []schema.Field {
 	return fields
 }
 
+func (enc *Encoder) writeListStart() {
```
zombiezen: These 3 helper functions make it harder to follow the logic below. Please inline them.

encoding/text/marshal.go
```diff
@@ -58,16 +58,22 @@ func (enc *Encoder) UseRegistry(reg *schemas.Registry) {
 	enc.nodes.UseRegistry(reg)
 }
 
+// SetIndent sets number of spaces to indent each level with.
+// A value of 0 disables indentation.
+func (enc *Encoder) SetIndent(indent int) {
```
zombiezen: Rather than setting an integer number of spaces, this could use a string, which would allow indents with tabs, too.

encoding/text/marshal.go
```diff
@@ -453,3 +544,65 @@ func (ew *errWriter) WriteByte(b byte) error {
...
+
+func (iw *indentWriter) Indent() {
+	iw.currentIndent++
+}
+
+func (iw *indentWriter) Unindent() {
+	iw.currentIndent--
+}
+
+func (iw *indentWriter) NewLine() {
```
zombiezen: I wonder whether the calling logic becomes simpler if `NewLine` and `NewLineOrSpace` were removed and the `indentWriter` would automatically apply this logic when a `\n` is encountered in its stream. WDYT?

encoding/text/marshal.go
martin-sucha: Done.

encoding/text/marshal_test.go
martin-sucha: Done.

encoding/text/marshal.go
martin-sucha: Done.

encoding/text/marshal.go
martin-sucha: Done.

encoding/text/marshal.go
martin-sucha: I think it would complicate the code instead of simplifying it. `indentWriter` would need to parse all the strings for `\n`, so the write couldn't just pass the buffer to the underlying writer. On top of that, we have `NewLine` which inserts a new line or nothing and `NewLineOrSpace` which inserts a new line or space. We need to distinguish those two cases, which would mean that we'd need to mark it somehow in the stream, complicating the parsing logic further. So I'd rather keep `NewLine` and `NewLineOrSpace` as they are now.

zombiezen: Acknowledged.

PR #146, opened 2019-08-05
Fix capnpc-go to use the correct length for nodes in the schema var.

kormat:
As there can only be `len(fnodes)` of nodes in a schema var, use that
for `NewNodes()` instead of the the number of files. This makes the
output deterministic for compiling a .capnp file on its own, or with
others on the same cmdline.

Fixes https://github.com/capnproto/go-capnproto2/issues/145

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -10,6 +10,7 @@
 
 # Please keep the list sorted.
 
+Anapaya Systems AG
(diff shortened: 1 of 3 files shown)


zombiezen:
Thank you for the fix!

PR #148, opened 2019-08-12
Fix wrong error check in schemas.go

### LINKED ISSUE #147: Wrong if condition in schemas.go
The following if condition
https://github.com/capnproto/go-capnproto2/blob/4a2d721f0c4b15f3e369dc4a82d604ef605d7da5/schemas/schemas.go#L114-L115
is wrong. The error is assigned to `r.err` but the check below is on `err` which was previously checked.
Not sure what the correct solution would be, probably the if should just check for `r.err`, or is the assignment to `r.err` wrong in the first place?

Found with go nilness analyzer (https://godoc.org/golang.org/x/tools/go/analysis/passes/nilness)

lukedirtwalker:
The error is assigned to r.err but checked is err, fix this by checking r.err.
Found with nilness analyzer.

Fixes #147

Also delete unreachable return nil in marshal.go

### DIFF
--- a/encoding/text/marshal.go
+++ b/encoding/text/marshal.go
@@ -461,7 +461,6 @@ func (enc *Encoder) marshalList(elem schema.Type, l capnp.List) error {
 	default:
 		return fmt.Errorf("unknown list type %v", elem.Which())
 	}
-	return nil
 }
 
 func (enc *Encoder) marshalEnum(typ uint64, val uint16) error {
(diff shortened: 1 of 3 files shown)


encoding/text/marshal.go
```diff
@@ -461,7 +461,6 @@ func (enc *Encoder) marshalList(elem schema.Type, l capnp.List) error {
 	default:
 		return fmt.Errorf("unknown list type %v", elem.Which())
 	}
-	return nil
```
zombiezen: Revert this line; unrelated?

lukedirtwalker: Yeah, I updated the commit message with the following sentence:
`Also delete unreachable return nil in marshal.go` I could also add this to the description or would you prefer a separate PR?

zombiezen: Makes sense. This is fine.

PR #157, opened 2020-08-24
Don't panic when Payload is null.

zenhack:
This patch fixes an issue where a null Payload in a Return or Call
message could cause the rpc logic to panic. Instead, we just treat a
null payload the same as an empty one.


zombiezen:
Thanks, Ian!

zenhack:
No problem!

Quoting Ross Light (2020-08-24 18:16:22)

PR #163, opened 2021-02-16
Typed refcounting

zenhack:
I've found myself wanting these in the past (particularly AddRef());
otherwise it's very un-ergonomic to have to do Foo{ f.Client.AddRef() }
at every call site.

---

Note that the first commit is from #162; this was the context in which I wanted those fixes. So you should probably review & merge that first.


zombiezen (approved):
LGTM. Please rebase and then I'll merge in.

zenhack:
Done

PR #168, opened 2021-05-12
Fix (*rpc.Conn).Err() method

Pr0Ger:
Right now this method will always return nil because until connection is closed (`c.state == connDead`) `c.closeErr` is nil.

This even contradicts with docstring which says `returns nil before Done is closed` when actually `.Done()` channel is closed after `c.state` became `connDead`

Proposed change will fix this


lthibault (approved):
LGTM

lthibault:
P.S. @Pr0Ger you may want to use the `v3` branch.  This is where active development is taking place.

Issue #166, opened 2021-03-04
Change import path to capnproto.org/go/capnp


lthibault:
Pinging @zombiezen @kentonv .

As part of https://github.com/capnproto/go-capnproto2/pull/165, it has become necessary to instantiate a go module around this project.  @taylorjdawson Initially tried to do the obvious, sensible thing and instantiate it with `go mod init github.com/capnproto/go-capnproto2`, with [predictable results](https://github.com/capnproto/go-capnproto2/pull/165#issuecomment-789926742) (at the risk of explaining the joke,  he should have used `zombiezen.com/go/capnproto2`).  This detail has briefly confused me on occasion as well, and I worry that it might frustrate new Go developers.

Additionally, many editors are configured to add a package alias when the last segment does not match the package name.  Ross [recently brought this up during code review](https://github.com/capnproto/go-capnproto2/pull/161#discussion_r575696053) and I ended up merging it anyway, after trying to figure out how to edit my VSCode configuration for 20 minutes without success. 😞 

After discussing this in slack with @zenhack and @taylorjdawson, we're of a mind to to change all the import paths in the project from `zombiezen.com/go/capnproto2` to (ideally) `capnproto.org/go/capnp`.

@zombiezen Is that okay with you?  In addition to asking if this might have technical consequences, I wanted to check that you were okay with this first because it's always a bit delicate when the new guys start rebranding something you built!

@kentonv Assuming ross is onboard with this, would you be willing to add the appropriate `<meta>` tag so that we could import this project as `capnproto.org/go/capnp`?

Miscellaneous thought:  I think this is probably safe to do since we're working towards a major version bump anyway.

kentonv:
Sorry, I don't know anything about Go imports. Can you explain what you mean about a `<meta>` tag?

On another note, I thought @zombiezen was working on a go-capnproto3 at one point, is that abandoned now?

lthibault:
> Sorry, I don't know anything about Go imports. Can you explain what you mean about a `<meta>` tag?

This is my fault.  I should have described this in terms of desired behavior instead of revealing my poor knowledge of DNS.  

What I mean is that we would like `capnproto.org/go/capnp` to resolve/redirect to `github.com/capnproto/go-capnproto2`.  This would allow us to install go-capnproto via `go get capnptoto.org/go/capnp`.  @zenhack suggested that this was commonly done via DNS `<meta>` tags, but I don't think the exact mechanism matters -- it's the behavior we're interested in.

[...]

PR #170, opened 2021-05-12
Update README.md for v3

lthibault:
I've taken the liberty of cleaning up the README a bit in preparation for `v3`.  In particular, certain details of the installation process are no longer relevant with the upcoming switch to modules (see: #165, #166).

Additionally, I've made some rather opinionated changes with respect to "marketing" (or -- if you prefer -- with respect to the ordering of information by priority).  I'm of the opinion that README's should first and foremost answer the question "what is this software, and is it for me?".  The answers to these questions are somewhat scattered  and buried below technical details, which IMHO only become relevant _after_ the initial what/why questions have been answered.

My basic approach can be summarized as follows:

1. In the heading, briefly what Capnp is, what go-capnp is, and why you might want to use it.
2. If you're still reading, your next question is likely to be "how do I get it"? ==> installation
3. If you successfully installed, your next question is "how do I start using it?"  ==> getting started
4. Once you have a minimal working example, you probably want to do something specific ==> API ref (godoc)
5. Once you've mapped the docs onto a proof-of-concept in your head, you're wondering if there are caveats/gotchas ==> API compatibility notice + license

Feedback very much welcome.

P.S.:  I added a couple of badges because nothing says "serious project" like badges.  Devs love badges. 😝

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,66 +1,55 @@
 # Cap'n Proto bindings for Go
 
-[![GoDoc](https://godoc.org/zombiezen.com/go/capnproto2?status.svg)][godoc]
-[![Build Status](https://travis-ci.org/capnproto/go-capnproto2.svg?branch=master)][travis]
+[![GoDoc](https://godoc.org/capnproto.org/go/capnp/v3?status.svg)][godoc]
+![License](https://img.shields.io/badge/license-MIT-brightgreen?style=flat-square)
+![tests](https://github.com/capnproto/go-capnproto2/workflows/Go/badge.svg)
 
-go-capnproto consists of:
-- a Go code generator for [Cap'n Proto](https://capnproto.org/)
-- a Go package that provides runtime support
-- a Go package that implements Level 1 of the RPC protocol
+[Cap’n Proto](https://capnproto.org/) is an insanely fast data interchange format similar to [Protocol Buffers](https://github.com/protocolbuffers/protobuf), but much faster.
 
-[godoc]: https://godoc.org/zombiezen.com/go/capnproto2
-[travis]: https://travis-ci.org/capnproto/go-capnproto2
+It also includes a sophisticated RPC system based on [Object Capabilities](https://en.wikipedia.org/wiki/Object-capability_model), ideal for secure, low-latency applications.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,66 +1,55 @@
 # Cap'n Proto bindings for Go
 
-[![GoDoc](https://godoc.org/zombiezen.com/go/capnproto2?status.svg)][godoc]
-[![Build Status](https://travis-ci.org/capnproto/go-capnproto2.svg?branch=master)][travis]
+[![GoDoc](https://godoc.org/capnproto.org/go/capnp/v3?status.svg)][godoc]
+![License](https://img.shields.io/badge/license-MIT-brightgreen?style=flat-square)
```
zenhack, line 4: Tangentially related, it appears GitHub's auto-detection of licenses can't work out which license we use, probably due to formatting oddities.

README.md
```diff
@@ -1,66 +1,55 @@
...
-- a Go package that provides runtime support
-- a Go package that implements Level 1 of the RPC protocol
+[Cap’n Proto](https://capnproto.org/) is an insanely fast data interchange format similar to [Protocol Buffers](https://github.com/protocolbuffers/protobuf), but much faster.
 
-[godoc]: https://godoc.org/zombiezen.com/go/capnproto2
-[travis]: https://travis-ci.org/capnproto/go-capnproto2
+It also includes a sophisticated RPC system based on [Object Capabilities](https://en.wikipedia.org/wiki/Object-capability_model), ideal for secure, low-latency applications.
 
-## Getting started
+### This package provides:
```
zenhack, line 11: Nitpick: jumping from an `<h1>` to an `<h3>` is a bit ugly. I don't think this necessarily needs to be a heading, even.

zenhack (approved):
Looks reasonable to me; a couple comments in-line but nothing I'd block over.

README.md
lthibault: What would be the course of action, here?

zenhack: We could futz with it, but I don't care that much, just something I noticed when I saw this.

Issue #178, opened 2021-07-24
First-time installation


taufik-rama:
Hello, I've been trying to to learn about proto serialization. I've been using `protocolbuffers` in my workplace & I want to know about the alternatives.

I have an issue during the installation. When I tried
```bash
$ go get capnproto.org/go/capnp/v3
```
I got an error
```bash
captain@glados:~$ go get capnproto.org/go/capnp/v3
cannot find package "capnproto.org/go/capnp/v3" in any of:
	/usr/local/go/src/capnproto.org/go/capnp/v3 (from $GOROOT)
	/home/captain/go/src/capnproto.org/go/capnp/v3 (from $GOPATH)
```
Same using `GO111MODULE` turned off
```
captain@glados:~$ GO111MODULE=off go get capnproto.org/go/capnp/v3
cannot find package "capnproto.org/go/capnp/v3" in any of:
	/usr/local/go/src/capnproto.org/go/capnp/v3 (from $GOROOT)
	/home/captain/go/src/capnproto.org/go/capnp/v3 (from $GOPATH)
```

My `go env` is
```
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/captain/.cache/go-build"
GOENV="/home/captain/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/captain/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/captain/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build046804836=/tmp/go-build -gno-record-gcc-switches"
[...]

PR #179, opened 2021-07-24
Added minor installation note regarding 'GO111MODULE'

taufik-rama:
Related to #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ This package provides:
 ## Installation
 
 ```
-$ go get capnproto.org/go/capnp/v3
+$ GO111MODULE=on go get capnproto.org/go/capnp/v3
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -22,6 +22,8 @@ $ go get capnproto.org/go/capnp/v3
 
 **NOTE:** You will need to install the [`capnp` tool](https://capnproto.org/capnp-tool.html) in order to compile your Cap'n Proto schemas into Go.  This package has been tested with version `0.8.0` of the `capnp` tool.
 
+Package might not be found due to how Go's module works, in which case you can try setting `GO111MODULE=on` for the command
```
lthibault, line 25: The installation procedure is getting kinda complex, so I think we should simplify and make this part of the default installation procedure.

Could you please integrate this as a (non-optional) part of the shell command?  In other words, change the `go get` line above to something like this ought to do the trick:

```bash
$ GO111MODULE=on go get capnproto.org/go/capnp/v3
```

lthibault (changes requested):

README.md
taufik-rama: Changed

lthibault (approved):

PR #181, opened 2021-10-21
Update the bundled schema to those included with capnp-c++ 0.9

zenhack:
Along the way, I automated the addition of the necessary $Go.name
annotations via a new "patch" subcommand to gen.sh (and a corresponding
patch), so that these won't need to be applied manually the next time
we update.

Additionally, the script needed a bit of tweaking to deal with the
fact that upstream now ships schema in subdirectories (namely compat/).


zenhack:
It means the compiler won't barf if you use the `-> stream` syntax, which is what prompted me to do this. But there are no runtime changes, and we don't act on the `-> stream` annotations in any way.

I'm not sure how how much of the C++ implementation's special treatment is even necessary; from an API perspective, the go implementation is already blocking, so in a sense we got there before the C++ implementation did, and I've had decent luck with writing the naive loop version of streaming as the C++ API changes mostly allowed you to do.

I don't fully understand the flow control issues involved; it's possible that some tweak to the way we queue calls internally could improve performance, but I don't think we even need the `-> stream` syntax for that, since again the API is already blocking.

lthibault:
Got it.  Thanks for the additional context 👍

PR #187, opened 2021-10-30
Improve error handling in RPC package

lthibault:
On several occasions I have found myself wanting to inspect RPC exceptions to control execution flow, in particular via functions such as `errors.Is` or `errors.As`.  Currently, errors are collapsed when composed, making them difficult to inspect.

This PR is an incremental attempt at improving error composition in the code base.  It exports the `errors.Error` type, enhances it with an `Unwrap()` method to make it compatible with the standard library `errors` package, and replaces `%v` formatting directives with `%w` and `fmt.Errorf`.

In addition to making errors more interpretable for the user, the resulting code is much more readable IMHO.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,6 @@ go 1.16
 
 require (
 	github.com/kylelemons/godebug v1.1.0
-	github.com/stretchr/testify v1.7.0 // indirect
+	github.com/stretchr/testify v1.7.0
 	github.com/tinylib/msgp v1.1.5
 )
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -31,7 +31,7 @@ func (e Error) Error() string {
 func (e Error) Unwrap() error { return e.Cause }
 
 func (e Error) GoString() string {
-	return fmt.Sprintf("errors.Error{Type: %s, Prefix: \"%s\", Cause: fmt.Errorf(\"%v\")}",
+	return fmt.Sprintf("errors.Error{Type: %s, Prefix: %q, Cause: fmt.Errorf(%q)}",
 		e.Type.GoString(),
 		e.Prefix,
 		e.Cause)
--- a/internal/errors/errors.go
+++ b/internal/errors/errors.go
@@ -9,9 +9,9 @@ import (
 
 // Error holds a Cap'n Proto exception.
 type Error struct {
-	Type   Type
-	Prefix string
-	Cause  error
+	ExcType Type
+	Prefix  string
+	Cause   error
 }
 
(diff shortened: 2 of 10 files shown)


internal/errors/errors.go
```diff
@@ -1,57 +1,76 @@
...
 
-func (e *capnpError) GoString() string {
-	return "errors.New(" + e.typ.GoString() + ", " + strconv.Quote(e.prefix) + ", " + strconv.Quote(e.msg) + ")"
+func (e Error) Unwrap() error { return e.Cause }
+
+func (e Error) GoString() string {
+	return fmt.Sprintf("errors.Error{Type: %s, Prefix: \"%s\", Cause: fmt.Errorf(\"%v\")}",
+		e.Type.GoString(),
+		e.Prefix,
+		e.Cause)
```
zenhack, line 37: Is there a reason these are no longer quoted?

go.mod
```diff
@@ -4,5 +4,6 @@ go 1.16
 
 require (
 	github.com/kylelemons/godebug v1.1.0
+	github.com/stretchr/testify v1.7.0
```
zenhack, line 7: I'm dubious about pulling in a dependency for something as trivial as what we're using this for here.

lthibault:
@zenhack I just rebased against `main` and double-checked that the errors string was properly formatted.  Are you ok to merge this?

internal/errors/errors.go
zenhack: Still waiting on a response to this comment; the old version properly escapes the strings, why was this changed?

lthibault:
@zenhack Re your question about quotes:  reason is twofold:

1. The `Sprintf` call made things readable enough IMO without the `strconv.Quote` call
2. Error message is still quoted (quotes are inline, negating the need for `strconv.Quote`)
3. An `error` value can't be passed to `strconv.Quote` directly.  Calling the `Error()` method seemed verbose.

tl;dr:  it was a judgement call about readability.  Outputs should be identical, though.

zenhack:
`strconv.Quote` is more than just `return "\"" + s "\""`; it escapes things inside the string as well.

But you can also use a `%q` format specifier, which would keep the full escaping while also being more concise. It also works with errors.

lthibault:
> `strconv.Quote` is more than just `return "\"" + s "\""`; it escapes things inside the string as well.

Oh, I didn't realize that.  Makes sense.

> But you can also use a `%q` format specifier, which would keep the full escaping while also being more concise. It also works with errors.

This sounds like our winner.  Will push changes shortly.

zenhack (approved):

PR #191, opened 2021-11-28
Add new internal/mpsc package.

zenhack:
Per discussion on matrix, we will need an unbounded queue implementation
to go forward with the deadlock fixes. Not currently hooked up to
anything, but the plan would be:

- Spawn a new goroutine per connection, which reads from the queue and
  writes messages to the transport until its context is canceled.
- Everywhere we currently call sendMessage(), instead insert something
  into this queue.

Note that the elements of the queue will not just be *capnp.Message,
becuase of the way the transport works; they may have to be callbacks or
somesuch.

---

@lthibault, note that since we agreed to go ahead with the testing library from #187, I decided to use it here as well, so these are going to conflict due to the double-add, but the conflicts will be easy to solve.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,5 +4,6 @@ go 1.16
 
 require (
 	github.com/kylelemons/godebug v1.1.0
+	github.com/stretchr/testify v1.7.0 // indirect
 	github.com/tinylib/msgp v1.1.5
 )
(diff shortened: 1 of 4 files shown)


lthibault (approved):
LGTM.

Issue #194, opened 2021-12-05
Documentation for server.Policy does not agree with implementation


zenhack:
The documentation for `server.Policy` indicates that when the limits are exceeded calls will immediately return exceptions. However, when pairing to debug #189, @lthibault and I discovered this is not actually what happens. Instead, the receive loop just blocks until an item in the queue clears.

However, I do not actually think what the documentation describes is what we want; this would mean lost messages if the sender is sending too fast, and it seems like it would be challenging to program around. We really want some kind of backpressure like that provided by the C++ implementation's streaming support.

I propose the following roadmap for dealing with memory/backpressure instead:

1. Short term, fix the docs to describe what the code actually does.
2. Slightly longer term, get rid of the bounded queue and replace it with an unbounded one.
3. Remove the `server.Policy` type entirely.
4. Implement something like the C++ implementation's per connection flow limit, for some soft backpressure at the connection level. Probably as a field on `rpc.Options`
5. Implement per-object flow control as is done in the C++ implementation for streaming specifically. I think we can do this for all clients irrespective of the streaming annotation, without changing any APIs; we just need to block the caller in appropriate places.
6. In `rpc.Options`, add a field for a hard limit on memory consumed by outstanding rpc messages (of all types, including returns), after which  the connection will just be dropped.

Thoughts?

@zombiezen, particularly since I'm proposing removing something you added for v3, I would appreciate your thoughts.

zombiezen:
> The documentation for `server.Policy` indicates that when the limits are exceeded calls will immediately return exceptions. [...] Instead, the receive loop just blocks until an item in the queue clears.

Yeah, looks like I introduced that in 27748fa54354994b866f4fb57f050e2bad34df8f without updating docs. Sorry.

> However, I do not actually think what the documentation describes is what we want; this would mean lost messages if the sender is sending too fast, and it seems like it would be challenging to program around. We really want some kind of backpressure like that provided by the C++ implementation's streaming support.

I agree, `server.Policy` seems like a kludge and it would be a better developer experience to have backpressure than to have a capability reject method calls. My memory is hazy here on details, but it seems like I might have realized the mistake and was trying to move more toward backpressure.

---

I did spill a lot of words at the time I was crunching through this that might help explain: https://groups.google.com/g/capnproto/c/A4LiXXd1t94/m/GkFH1AK-AgAJ

[...]

PR #196, opened 2021-12-06
Fix RPC deadlock

### LINKED ISSUE #190: Possible deadlock when receive loops on both sides of an unbuffered connection try to send.
While debugging #189, @lthibault and I realized there is another possible deadlock bug which was not triggered by his test program, but which is theoretically possible:

If two return messages are in being processed at the same time, one on each side of a connection (where both implementations are go-capnp, as opposed to some other language), each will try to send a correspdonding Finish message from within the receive loop, and it will not process any further messages until this finish has been written to the connection.

If the connection is unbuffered (e.g. net.Pipe), this could cause a deadlock, as neither side's send will complete until the other side tries to receive.

I suspect there may be other similar bugs. As a general principle, we should endeavor for the receive loop to never block, and I would be surprised if these were the only deadlock bugs that resulted from violating this general principle. (The cause of #189 itself is more complicated, but also involves blocked receive loops).

lthibault:
Fixes #190 and #194.

@zenhack This is still WIP, but I want to get some early feedback.  Can you look over this and make sure I'm on the right track?  Thanks.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -164,13 +164,8 @@ func NewConn(t Transport, opts *Options) *Conn {
 	}
 
 	// start background tasks
-	for _, f := range []func(context.Context) error{
-		c.send,
-		c.receive,
-	} {
-		c.tasks.Add(1)
-		g.Go(c.newTask(f))
-	}
+	g.Go(c.newTask(c.send))
+	g.Go(c.newTask(c.receive))
 
 	// monitor background tasks
 	go func() {
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -10,6 +10,7 @@ import (
 	"capnproto.org/go/capnp/v3"
 	"capnproto.org/go/capnp/v3/internal/errors"
 	rpccp "capnproto.org/go/capnp/v3/std/capnp/rpc"
+	"golang.org/x/sync/errgroup"
 )
 
 /*
@@ -67,7 +68,7 @@ is a common source of errors and/or inefficiencies.
 // It is safe to use from multiple goroutines.
 type Conn struct {
 	bootstrap    *capnp.Client
-	reporter     ErrorReporter
+	er           annotatingErrReporter
 	abortTimeout time.Duration
 
 	// bgctx is a Context that is canceled when shutdown starts.
@@ -93,6 +94,8 @@ type Conn struct {
 	// progress, and the channel is closed when the operation is finished.
 	// See the above comment for a longer explanation.
 	sendCond chan struct{}
+
+	sendq *sendQueue
 
 	// Tables
 	questions  []*question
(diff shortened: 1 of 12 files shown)


rpc/rpc.go
```diff
@@ -137,44 +140,60 @@ type ErrorReporter interface {
...
-	c.tasks.Add(1)
+
+	// start background tasks
+	for _, f := range []func(context.Context) error{
+		c.send,
+		c.receive,
+	} {
+		c.tasks.Add(1)
+		g.Go(c.newTask(f))
+	}
```
zenhack, line 173: The loop feels very over-engineered. I would say:

- Move the call to `Add` into the body of newTask
- Change this to just:

```go
c.Go(c.newTask(c.send))
c.Go(c.newTask(c.receive))
```

The loop might make sense for some N > 2, but at this size I think it just makes the code obscure for comparatively little benefit.

rpc/send.go
```diff
@@ -0,0 +1,108 @@
...
+
+type errChan chan error
+
+func (cherr errChan) ReportError(err error) {
+	cherr <- err // buffered
+}
+
+var cherrPool = make(errChanPool, 64)
+
+type errChanPool chan errChan
```
zenhack, line 83: I don't really understand why this was introduced? What is this doing for you?

rpc/send.go
```diff
@@ -0,0 +1,108 @@
...
+	})
+}
+
+func (sq *sendQueue) Recv(ctx context.Context) preparer {
+	v, err := (*mpsc.Queue)(sq).Recv(ctx)
+	if err != nil {
+		return sendFailure(err)
+	}
+
+	return *(*preparer)(unsafe.Pointer(&v))
```
zenhack, line 47: I don't think this is correct; the inserted value is a sendReq, and I'm not sure the memory layout will be the same as if it is passed in as a preparer.

You should probably just use a type assertion here anyway:

```go
v.(sendReq)
```

...I do not think `unsafe`is warranted here.

rpc/send.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+type prepFunc func(rpccp.Message) error
+
+func (prepare prepFunc) Prepare(m rpccp.Message) error { return prepare(m) }
+
+func sendFailure(err error) prepFunc {
+	return func(rpccp.Message) error { return err }
```
zenhack, line 55: There is a lot of indirection in this file that as far as I can tell only serves to simplify the body of `Conn.send`. But my feeling is it would probably be better to have some if/elses in there than to do it this way; this seems very convoluted.

zenhack (changes requested):
This seems like it is broadly on the right track, though some parts I have only skimmed. A few things mentioned in-line.

lthibault:
I think it's going to be easier to start again from scratch.  Closing.

PR #198, opened 2021-12-11
Flow control

zenhack:
Still much to do here (including docs), but I have started working on support for per-object flow control. The plan is to add a `SetFlowLimiter` method to `Client`. Possibly after this is merged we can also add a `FlowLimiter` that is adaptive, based on BBR or the like, rather than just using a fixed window.

### DIFF
--- a/rpc/flow_test.go
+++ b/rpc/flow_test.go
@@ -5,6 +5,7 @@ package rpc
 import (
 	"context"
 	"net"
+	"os"
 	"sync"
 	"testing"
 	"time"
@@ -61,6 +62,12 @@ func (t *measuringTransport) RecvMessage(ctx context.Context) (rpccp.Message, ca
 // and then send calls as fast as the limiter will let us. At the end, we check
 // that the maximum size of outstanding messages looks right.
 func TestFixedFlowLimit(t *testing.T) {
+	if os.Getenv("FLAKY_TESTS") != "1" {
+		t.Skip("Not running TestFlowFixedLimit, which is flaky. Set FLAKY_TESTS=1 to enable")
+		// TODO: at some point make this test run robustly in CI;
+		// it seems to work reliably when run locally for both @zenhack
+		// and @lthibault
+	}
 	t.Parallel()
 
 	limit := uint64(1 << 20)
--- a/rpc/flow_test.go
+++ b/rpc/flow_test.go
@@ -1,0 +1,151 @@
+package rpc
+
+// tests for streaming/flow control
+
+import (
+	"context"
+	"net"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"capnproto.org/go/capnp/v3"
+	"capnproto.org/go/capnp/v3/flowcontrol"
+	"capnproto.org/go/capnp/v3/rpc/internal/testcapnp"
+	"capnproto.org/go/capnp/v3/server"
+	rpccp "capnproto.org/go/capnp/v3/std/capnp/rpc"
+)
+
+// Transport, which is a wrapper around another transport, and measures the total size of
+// all messages received from RecvMessage(), but not released. It tracks the current and
+// all-time-maximum of this value.
+type measuringTransport struct {
+	Transport
+
+	mu              sync.Mutex
+	inUse, maxInUse uint64
+}
+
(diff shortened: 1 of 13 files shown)


zenhack:
Alright, this is finished & tested, PTAL.

As a separate PR, at some point, we should provide an adaptive implementation of FlowLimiter, but this gets the machinery in place and at least lets you do flow control if you want to tell the library what the right size is.

zenhack:
Ok, looks like I need to fuss with the test a bit more; it passes on my machine, so perhaps I'm just not providing enough slack for github CI. I'll have to play with it so that it's loose enough to not tigger false positives, but not so loose that it doesn't catch real bugs.

Annoyingly, this kind of thing is sortof inherently racy, so it's a bit of an art figuring out how much slack is enough...

zenhack:
(In the meantime, this is still worth reviewing)

rpc/flow_test.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/stretchr/testify/assert"
+
+	"capnproto.org/go/capnp/v3"
+	"capnproto.org/go/capnp/v3/flowcontrol"
+	"capnproto.org/go/capnp/v3/rpc/internal/testcapnp"
+	"capnproto.org/go/capnp/v3/server"
+	rpccp "capnproto.org/go/capnp/v3/std/capnp/rpc"
+)
+
+// Transport, which is a wrapper around another transport, and measures the total size of
```
lthibault, line 21: `measuringTransport`

lthibault (approved):

zenhack:
@lthibault can you check out this branch and tell me what the new tests do on your end? Also open to specific ideas wrt. how to make the failing test more robust.

One thing I noticed when poking around: allocating a slot for the outstanding call on the server is a linear time operation. In the test I cranked `MaxConcurrentCalls` *way* up to avoid a situation where we weren't hitting the flow limit because the receive loop was blocking on a full server queue. But, this could result in actually spending a lot of CPU time in that allocation loop, and I wonder if in an environment with less available CPU that could still be a bottleneck. The CI run for 3f3cc1c reported notably lower flow on Go 1.15 than Go 1.17, which, assuming the compiler gets better over time, supports the idea that this may be a CPU bound problem.

Because of that, unless you have any great insights, or running it on your end leads to some discovery, I wonder if it wouldn't make sense to table the test until we've gotten rid of that queue, since we're planning on doing that anyway, and then pick it up afterwards. I think I don't want to leave the PR hanging though, since it touches some bits of the code I want to work on for #200; I suppose we could leave it in with a `t.Skip(...)` for now...

lthibault:
@zenhack 

Seems to be working fine for me.  Though did you perhaps remove the offending test?  `TestFixedFlowLimit` is the test that fails, but I only see `TestFixed` in the current tip of your branch (presumably, this is just a renaming).

>this could result in actually spending a lot of CPU time in that allocation loop

This would be `fixedLimiter.pumpQueue`, yes?

Perhaps it would be more performant to use [golang.org/x/sync/semaphore](https://pkg.go.dev/golang.org/x/sync/semaphore), here?  This avoids the busy loop, which is presumably what's burning up CPU cycles.

>I wonder if it wouldn't make sense to table the test until we've gotten rid of that queue
> I think I don't want to leave the PR hanging though

Yes, this sounds reasonable to me 👍 

```
$ go test -v ./flowcontrol/...
=== RUN   TestFixed
--- PASS: TestFixed (0.02s)
PASS
ok      capnproto.org/go/capnp/v3/flowcontrol   0.191s
``` 

```
$ go test -v -race ./flowcontrol/...
=== RUN   TestFixed
--- PASS: TestFixed (0.02s)
PASS
ok      capnproto.org/go/capnp/v3/flowcontrol   0.200s
```

zenhack:
Quoting Louis Thibault (2021-12-19 14:57:18)

>    Seems to be working fine for me. Though did you perhaps remove the
>    offending test? TestFixedFlowLimit is the test that fails, but I only
>    see TestFixed in the current tip of your branch (presumably, this is
>    just a renaming).

No, these are two different tests and they should both be there. The
failing test is in rpc/, not flowcontrol/.

>    This would be fixedLimiter.pumpQueue, yes?

No, I'm talking about Server.nextID in server/server.go, which is pre-existing
code.

zenhack:
Disabled the test, per discussion on matrix.

zenhack:
@lthibault, are you comfortable merging this?

lthibault (approved):

Issue #189, opened 2021-11-15
Non-deterministic deadlock on concurrent RPC calls


lthibault:
As discussed in the [Matrix channel](https://matrix.to/#/!pLcnVUHHRZrUPscloW:matrix.org/$asvpMbN2GCVMQZsgNbN0r13hkSaUrlmckHlF87TRS30?via=matrix.org), concurrent RPC calls seem to provoke an intermittent deadlock.

I am attaching [scratch.zip](https://github.com/capnproto/go-capnproto2/files/7540535/scratch.zip), which contains a working example to reproduce this behavior.   The example was built on commit `e33fd9381f266af736f49ab7366c49f0397e3c16` of go-capnproto2 and version 0.9.1 of the capnp tool.

Summarizing a few key points from the Matrix discussion:

1. This bug also manifests using out-of-process transports, notably TCP to localhost.
2. The deadlock appears to occur when n>2 concurrent calls are made
    - This suggests a full buffer/channel somewhere
    - The synchronous semantics of `net.Pipe` likely cause this to happen sooner

@zenhack IFF you have the time, I would be super interested in pair-programming to debug this.  This seems like a good opportunity to get more involved in the internals.

PR #204, opened 2021-12-17
Creating example directory with books and hashes examples from wiki

frumioj:


example/hashes/hashes/hashes.capnp
```diff
@@ -0,0 +1,13 @@
...
+$Go.import("foo/hashes");
+
+interface HashFactory {
+    newSha1 @0 () -> (hash :Hash);
+}
+
+interface Hash {
+    write @0 (data :Data) -> ();
+    sum @1 () -> (hash :Data);
+}
```
lthibault, line 13: missing newline

example/books/books.capnp
```diff
@@ -0,0 +1,12 @@
...
+$Go.package("books");
+$Go.import("foo/books");
+
+struct Book {
+    title @0 :Text;
+    # Title of the book.
+
+    pageCount @1 :Int32;
+    # Number of pages in the book.
+}
```
lthibault, line 12: missing newline.

lthibault (changes requested):
John, thanks a million for this contribution!  I've been meaning to do something similar, but never quite found the time.  This really helps a lot! 🙂 

I've flagged a couple of minor formatting issues.  Otherwise LGTM.

(EDIT:  I did note that this is currently a copy/paste of the non-building examples.  But I appreciate the initiative.  We'll get this fixed!)

zenhack:
I opened a pr on this pr: https://github.com/frumioj/go-capnproto2/pull/1

Can confirm that after fixing the build issues it just hangs. Here's an interesting observation: If I apply this additional patch, it works:

```
diff --git a/example/hashes/hashesserver.go b/example/hashes/hashesserver.go
index f09b444..24ba1c1 100644
--- a/example/hashes/hashesserver.go
+++ b/example/hashes/hashesserver.go
@@ -124,8 +124,25 @@ func client(ctx context.Context, rwc io.ReadWriteCloser) error {
 }
 
 func main() {
+	l, err := net.Listen("unix", "/tmp/fff")
+	chkfatal(err)
+
 	ctx := context.Background()
-	c1, c2 := net.Pipe()
-	go serveHash(ctx, c1)
+
+	go func() {
+		c1, err := l.Accept()
+		chkfatal(err)
+		serveHash(ctx, c1)
+	}()
+
+	c2, err := net.Dial("unix", "/tmp/fff")
+	chkfatal(err)
+
 	client(ctx, c2)
 }
+
+func chkfatal(err error) {
+	if err != nil {
+		panic(err)
+	}
+}
```

i.e. it works fine if you use a real network connection, instead of `net.Pipe()`. So I suspect this example is suffering from #189.

lthibault:
> i.e. it works fine if you use a real network connection, instead of `net.Pipe()`. So I suspect this example is suffering from #189.

@frumioj  The current work-around for this is to increase the `MaxConcurrentCalls` field in [`server.Policy`](https://pkg.go.dev/capnproto.org/go/capnp/v3@v3.0.0-alpha.1/server#Policy).  Note that this will deadlock if this number of concurrent calls is exceeded, so you'll need to set it to a sufficiently high value (and maybe track client-side connections)?

We're working on a fix for this.  ETA unknown, but it's basically our highest-priority item.

example/hashes/hashes/hashes.capnp
frumioj: Change accepted.

lthibault:
Is this ready to merge?

frumioj:
> Is this ready to merge?

I had planned to:
1. Get the books example done
2. Put documentation comments in my code

before merge. Can also do either/both as a new PR? WDYT?

zenhack:
Let's wait on these tasks before merging; it's not like we're likely to
have conflicts with other ongoing work.

Quoting John Kemp (2021-12-18 13:22:55)

frumioj:
In the books example code, I get a panic when reading the books schema into the decoder: `panic: capnp: decode: too many segments to decode` - when looking at the code that makes the error, I find it difficult to believe that this file results in > 512 segments? There is a newline at the end of the file now. Is this the correct way to call the exe?

```
johnk@REGEN:~/src/go-capnproto2/example/books/ex2$ cat ../books/books.capnp
@0x85d3acc39d94e0f8;
$Go.package("books");
$Go.import("foo/books");

struct Book {
    title @0 :Text;
    # Title of the book.

    pageCount @1 :Int32;
    # Number of pages in the book.
}
johnk@REGEN:~/src/go-capnproto2/example/books/ex2$ cat ../books/books.capnp | ./bookstest2 
panic: capnp: decode: too many segments to decode

goroutine 1 [running]:
main.main()
	/home/johnk/src/go-capnproto2/example/books/ex2/books2.go:15 +0x266
johnk@REGEN:~/src/go-capnproto2/example/books/ex2$ cat books2.go
package main

import (
    "fmt"
    "os"

    "books"
    "capnproto.org/go/capnp/v3"
)

func main() {
    // Read the message from stdin.
    msg, err := capnp.NewDecoder(os.Stdin).Decode()
    if err != nil {
        panic(err)
    }

    // Extract the root struct from the message.
    book, err := books.ReadRootBook(msg)
    if err != nil {
        panic(err)
    }

    // Access fields from the struct.
    title, err := book.Title()
    if err != nil {
        panic(err)
    }
    pageCount := book.PageCount()
    fmt.Printf("%q has %d pages\n", title, pageCount)
}
```

zenhack:
The example program is expecting to read a capnproto encoded message from stdin, containing a `Book` struct. Instead, you're feeding it the text of the book schema. So it's interpreting the first 4 bytes of that as the segment count for the message, and ascii `@0x8` interpreted as a little-endian integer is indeed greater than 512.

If you want to construct messages from text for testing purposes, look at the `capnp` tool's `convert` and `eval` commands.

example/books/ex1/.gitignore
```diff
@@ -0,0 +1 @@
+/bookstest1
```
lthibault, line 1: Newline 😉

example/books/ex2/.gitignore
```diff
@@ -0,0 +1 @@
+/bookstest2
```
lthibault, line 1: Newline

frumioj:
> The example program is expecting to read a capnproto encoded message from stdin, containing a `Book` struct. Instead, you're feeding it the text of the book schema. So it's interpreting the first 4 bytes of that as the segment count for the message, and ascii `@0x8` interpreted as a little-endian integer is indeed greater than 512.
> 
> If you want to construct messages from text for testing purposes, look at the `capnp` tool's `convert` and `eval` commands.

Thanks. I created a `book.txt` with a cleartext message, and included instruction for creating the encoded value and reading it into the ex2 books example. I think it's getting pretty close!

frumioj:
Hi, I think I addressed the review changes (review is also outdated) but github prevents this from being merged. Is this PR OK @lthibault, or what else remains?

example/books/README.md
```diff
@@ -0,0 +1,14 @@
...
+```
+cd $GOPATH/src/go-capnproto2/example/books
+capnp compile -I/home/johnk/go/src/capnproto.org/go/capnp/std/ -ogo books/books.capnp
+```
+
+Then build and run each example:
+```
+cd ex1 && go build . && ./bookstest1
+
+cd ex2 && go build . && capnp encode ../books/books.capnp Book < ./book.txt > ./book.capnpenc && ./bookstest2 < ./book.capnpenc && cd ..
```
zenhack, line 13: Minor thing, rather than creating a temporary file, you could just use a pipe:

```suggestion
cd ex2 && go build . && capnp encode ../books/books.capnp Book < ./book.txt | ./bookstest2 && cd ..
```

example/books/books/books.capnp.go
```diff
@@ -0,0 +1,103 @@
+// Code generated by capnpc-go. DO NOT EDIT.
```
zenhack, line 1: Since this isn't really intended for people to fetch with go-get, maybe we should just not commit this to the repo? I generally dislike having build artifacts stored in git, and it seems like we don't actually need it here. Same goes for the other example(s).

Also, you comitted book.capnpenc, which we *definitely* don't need to version.

zenhack (changes requested):
This looks mostly reasonable, a couple comments in-line.

The one other thing I would suggest is that we at least build these in CI (if not actually run them), so we'll notice if the examples get out of date again.

Thanks for the patch!

example/books/books/books.capnp.go
frumioj: Yeah - good question. I did commit this generated file because in one OSS community I work in (Cosmos SDK) they have a requirement to commit the go files generated from their protocol buffers. Since the code that calls this generated file, depends on it, I could totally agree with that decision and leave it checked-in. Or... since it is generated from the .capnp, I could also remove it (which is what I would have done before I heard of projects that wanted the generated files committed). As maintainers, you might have to make a decision... I don't mind either way. I will remove book.capnpenc, although that is committed for the same reason... so I'll maybe also remove the generated go files too. Sound OK?

frumioj (commented):
LGTM

zenhack (dismissed):

zenhack:
LGTM. I'll wait for CI to run and @lthibault to sign off before merging.

example/books/ex1/books1.go
```diff
@@ -0,0 +1,30 @@
...
+	if err != nil {
+		panic(err)
+	}
+
+	// Create a new Book struct.  Every message must have a root struct.
+	book, err := books.NewRootBook(seg)
+	if err != nil {
+		panic(err)
+	}
+	book.SetTitle("War and Peace")
```
lthibault, line 22: nit:  String setters return an error.  It's totally okay to ignore it for a demo, but I think we should signal to readers that the error exists, and should be handled in production code.

Suggest:  `_ = book.SetTitle("War and Peace")`

example/hashes/hashesserver.go
```diff
@@ -0,0 +1,162 @@
...
+	"crypto/sha1"
+	"fmt"
+	"hash"
+	"io"
+	"net"
+	"os"
+
+	"capnproto.org/go/capnp/v3/rpc"
+	"capnproto.org/go/capnp/v3/server"
+	"golang.org/x/net/context"
```
lthibault, line 13: Shouldn't this just be the `"context"`?

lthibault (approved):
LGTM.

lthibault:
Bump @frumioj

frumioj:
> Bump @frumioj

Apologies - I somehow missed these comments (may have been on holiday I guess?) - will address ASAP!

example/books/ex1/books1.go
frumioj: Fixed in 48ca3918cc0b56e42c5119d884edbbf98a26f2c3

example/hashes/hashesserver.go
frumioj: fixed in 84a7de0a116bd09eb7487daf16cea767f7d94403

lthibault (approved):
LGTM  🎉

lthibault:
Many thanks once again for your help, John!

Out of curiosity, what are you building?

frumioj:
I'm working on building a distributed cryptographic key management system (https://github.com/frumioj/capnkeys). I'd like to use capnproto for the interface to that if possible.

lthibault:
Wow, that's serendipitous!  Believe it or not, I have a need for such a thing for [Wetware](https://github.com/wetware/ww/tree/refactor/casm).

Are you able to chat on Matrix by any chance?  You can find me [here](https://matrix.to/#/@lthibault:matrix.org).

PR #207, opened 2021-12-31
Cleanup: Stop passing around a superfluous []ClientState

zenhack:
This only actually gets used in sendCap(), but there we have a reference
to the corresponding client, so we can just call client.State() instead
of carrying around this parallel slice everywhere.


lthibault (approved):
Always happy to see code being removed! 🙂

PR #208, opened 2022-01-01
Support arbitrary client metadata, avoid linear seach in sendCap.

zenhack:
This solves an issue I discovered where sendCap() does a linear search
on the exports table for each sent capability.

After a long discussion with Louis over matrix, we concluded that adding
a metadata object for clients was the best solution; there's a general
design awkwardness right now where the rpc system needs to be able to
inspect the client in various ways, but Client is defined in the main
package and tries to encapsulate its implementation details. To some
degree we are able to use Brand(), but this gets awkward quickly,
because:

1. ClientHook is an openly implementable interface, so we don't know
   all the posibilities in the rpc code.
2. As I've been working on supporting receiverAnswer, I've found
   pinning down all the implementations I defined in the library that
   need to be worried about is challenging.
3. Some of these implementations aren't even exported, or just return
   nil from Brand().

The Metadata object lets the rpc system manage the bookkeeping
information independently of the internals of a client.

I expect to use this for more in the future, but for now this lets
us avoid the linear search by just storing the export id for each
connection in the client's metadata.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -464,6 +464,9 @@ type ClientState struct {
 	// Arbitrary metadata. Note that, if a Client is a promise,
 	// when it resolves its metadata will be replaced with that
 	// of its resolution.
+	//
+	// TODO: this might change before the v3 API is stabilized;
+	// we are not sure the above is the correct semantics.
 	Metadata *Metadata
 }
 
--- a/capability.go
+++ b/capability.go
@@ -139,8 +139,8 @@ func NewClient(hook ClientHook) *Client {
 		done:       make(chan struct{}),
 		refs:       1,
 		resolved:   newClosedSignal(),
+		metadata:   *NewMetadata(),
 	}
-	(&h.metadata).init()
 	h.resolvedHook = h
 	c := &Client{h: h}
 	if clientLeakFunc != nil {
@@ -167,8 +167,8 @@ func NewPromisedClient(hook ClientHook) (*Client, *ClientPromise) {
 		done:       make(chan struct{}),
 		refs:       1,
 		resolved:   make(chan struct{}),
+		metadata:   *NewMetadata(),
 	}
-	(&h.metadata).init()
 	c := &Client{h: h}
 	if clientLeakFunc != nil {
 		c.creatorFunc = 2
@@ -854,8 +854,8 @@ func ErrorClient(e error) *Client {
 		done:       make(chan struct{}),
 		refs:       1,
 		resolved:   newClosedSignal(),
+		metadata:   *NewMetadata(),
 	}
-	(&h.metadata).init()
 	h.resolvedHook = h
 	return &Client{h: h}
 }
--- a/capability.go
+++ b/capability.go
@@ -109,6 +109,9 @@ type clientHook struct {
 	// a clientHook.
 	ClientHook
 
+	// Place for callers to attach arbitrary metadata to the client.
+	metadata Metadata
+
 	// done is closed when refs == 0 and calls == 0.
 	done chan struct{}
 
@@ -136,6 +139,7 @@ func NewClient(hook ClientHook) *Client {
 		refs:       1,
 		resolved:   newClosedSignal(),
 	}
+	(&h.metadata).init()
 	h.resolvedHook = h
 	c := &Client{h: h}
 	if clientLeakFunc != nil {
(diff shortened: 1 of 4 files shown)


capability.go
```diff
@@ -455,6 +461,10 @@ type ClientState struct {
 	Brand Brand
 	// IsPromise is true if the client has not resolved yet.
 	IsPromise bool
+	// Arbitrary metadata. Note that, if a Client is a promise,
+	// when it resolves its metadata will be replaced with that
+	// of its resolution.
```
zenhack, line 466: Note that I'm not 100% sure these semantics are actually what is desired, or at least it seems like we might want to still be able to access the un-resolved version of this. `State()` is supposed to return a snapshot, so perhaps we also want `OriginalState()` which would recover pre-resolution state (in which case if `OrigState().IsPromise == false` then `State() == OrigState()`)?

...I feel like I don't totally understand the broader use of this, and I kindof expect whatever decisions we make here we'll likely discover we need to tweak exactly how this works a bit before stabilizing.

capability.go
```diff
@@ -136,6 +139,7 @@ func NewClient(hook ClientHook) *Client {
 		refs:       1,
 		resolved:   newClosedSignal(),
 	}
+	(&h.metadata).init()
```
lthibault, line 142: This more of a design/devex issue than anything else, but I feel that this should be handled by a constructor.  As it is, it adds visual noise to the code-base and raises the possibility of forgetting to call `init()`. 

In fact, I think `clientHook` should probably have a corresponding `newClientHook` function, but we don't have to address that in this PR.

capability.go
lthibault: 👍  Copy that.

Perhaps consider adding a `TODO` comment?

rpc/export.go
```diff
@@ -95,26 +122,28 @@ func (c *Conn) sendCap(d rpccp.CapDescriptor, client *capnp.Client, state capnp.
...
+		return id, true
 	}
+
+	// Not already present; allocate an export id for it:
 	ee := &expent{
 		client:   client.AddRef(),
 		wireRefs: 1,
 	}
-	id := exportID(c.exportID.next())
+	id = exportID(c.exportID.next())
```
lthibault, line 140: Shouldn't `next()` just return a correctly-typed `exportID`?  Again, visual noise.

lthibault (dismissed):
Approving this because it looks correct.  Nevertheless, I feel moderately strongly that the two readability issues I've flagged should be addressed before merging.  Your call.

lthibault:
Could [this check](https://github.com/capnproto/go-capnproto2/pull/208/files#diff-b3e7d47e460625492aa4e5f0d42b8195cd9179ab8f75def101604256ada2825bR115-R120) in `sendCap` be refactored to use `isLocalClient` from #212 ?

That would be a huge readability win.  The above snippet is quite jarring every time I look at it, despite the fact that the domain logic it represents is dead-simple.

rpc/export.go
zenhack: It was like that when I found it :P. But I think the reason is that `c.exportID` is of type `idgen`, which is used for generating other types of identifiers. Once generics lands, we could consider making idgen generic and changing this to be an `idgen(exportID)`. For now though, I think the code re-use outweights needing to cast.

capability.go
zenhack: Done.

capability.go
zenhack: The reason I did it this way is I wanted to have the `Metadata` live directly in the `clientHook`. There is `NewMetadata`, but it returns a pointer...

zenhack:
> Could this check in sendCap be refactored to use isLocalClient from #212 ?

I assume you meant to link to a specific line? The link just points to the whole diff, so I'm not sure which check you mean.

capability.go
lthibault: That's what I gathered, but couldn't we nevertheless return a concrete struct from the constructor?  It would copy the mutex, but there would no longer be any copies after first use.  Unless I'm misremembering, the Go toolchain should be perfectly happy with that.

lthibault:
Looks like it was merged, so no longer appearing in the diff:  https://github.com/capnproto/go-capnproto2/blob/8c878e053da7da0aa3d7c46f9c60c1338efaa84a/rpc/export.go#L88-L93

However, I just saw that we actually use the concrete `*importClient`, so never mind.

capability.go
zenhack: Ok, changed (I also tweaked the wording on the comment so it's clear it's ok to copy before first use, which addresses my main hangup).

I also merged in main and used syncutil.With in a couple places where it's appropriate.

lthibault (dismissed):
Cheers, much nicer to look at :)

lthibault (approved):

PR #209, opened 2022-01-01
Handle incoming `receiverAnswer` cap descriptors.

zenhack:
Marking this as a draft, still TODO:

- [x] Some more tests. In particular, the current test only excercises bootstrap messages, we should of course test calls as well.
- [x] Scrutinize the logic in recvCap for choosing when to return local = true; This is subtle and I'm too tired to think it through carefully right now, will get back to it.

Note that this includes the changes from #207 and #208, so the diff is bigger than it should be.


zenhack:
Ok, I think the logic for local in recvCap is right now, though I still need to write some more tests. Relatedly, this also includes #212, which should be merged first.

zenhack:
Alright, added another test, which projects on calls. I consider this ready to go, but these should be reviewed & merged first:

- #207 
- #208 
- #212

rpc/answer.go
```diff
@@ -202,13 +207,20 @@ func (ans *answer) Return(e error) {
...
 // The caller must be holding onto ans.c.mu and the sender lock.
 // The result's capability table must have been extracted into
-// ans.resultsCapTable before calling sendReturn. Only one of
+// ans.resultCapTable before calling sendReturn. Only one of
 // sendReturn or sendException should be called.
-func (ans *answer) sendReturn(cstates []capnp.ClientState) (releaseList, error) {
+func (ans *answer) sendReturn() (releaseList, error) {
 	ans.pcall = nil
 	ans.flags |= resultsReady
+	// XXX: do we need to check if we've received a finish? if so, the
```
lthibault, line 215: The `XXX` suggests that this is a piece of missing implementation.  Is that (still) the case?

lthibault (approved):
LGTM.  Let's maybe merge #208 first?

rpc/answer.go
zenhack: Good catch, fixed. Also fixed merge conflicts.

lthibault (approved):

PR #210, opened 2022-01-02
Add some helpers for acquiring locks

zenhack:
This adds a couple helpers "With" and "Without", which  make it easier
to follow where the boundaries of critical sections (and the holes in
them) are, when the helpers are applicable.

This is especially helpful in that using it consistently means places
where the lock management is *not* just a simple straight-line critical
section stand out.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -6,6 +6,7 @@ import (
 	"sync"
 
 	"capnproto.org/go/capnp/v3/internal/errors"
+	"capnproto.org/go/capnp/v3/internal/syncutil"
 )
 
 // A Promise holds the result of an RPC call.  Only one of Fulfill,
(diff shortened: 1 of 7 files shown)


answer.go
```diff
@@ -193,19 +194,19 @@ func (p *Promise) resolve(r Ptr, e error) {
...
 			p.callsStopped = make(chan struct{})
 		}
-		p.mu.Unlock()
-		res := resolution{p.method, r, e}
-		for path, row := range p.clients {
-			t := path.transform()
-			for i := range row {
-				row[i].promise.Fulfill(res.client(t))
-				row[i].promise = nil
+		syncutil.Without(&p.mu, func() {
```
lthibault, line 197: Just a small idea/suggestion:

I have a small library called [`github.com/lthibault/util/sync`](https://github.com/lthibault/util/sync) into which these functions would fit quite nicely.  By happy coincidence, the package name is also `syncutil`.  It could help declutter things a bit if we were to move it there and import the library.

lthibault (approved):
Love this.  In fact, I independently had the same idea a few months ago and never got around to implementing it.

Needless to say, approved.

answer.go
zenhack: Makes sense to me. I'm going to hit merge; feel free to move these and submit a pr.

It might also make sense to pull some other random packages out of `internal`, e.g. `mpsc`, `chanmutex`...

I am vaguely bothered by the fact that your package path doesn't agree with the package name, but meh.

zenhack:
Ah, I see the reason you didn't merge is because there were conflicts. Fixed.

answer.go
lthibault: >I am vaguely bothered by the fact that your package path doesn't agree with the package name, but meh.

I used to be the same way, but the `import project "github.com/user/project/pkg"` convention is quite popular, so I don't think it's "wrong" in Go.

The alternative was a name collision with the standard library `sync`.

Anyway 👍 .  Sounds like a plan.  Merge it!

PR #211, opened 2022-01-02
Expose pipelineClient & do something useful with its .Brand()

zenhack:
We will almost certainly need to reflect on pipelineClient with .Brand() in the
rpc system at some point. This patch:

- exports the type,
- adds a couple accessor methods, and
- changes .Brand() to return the PipelineClient in the non-resolved
  case.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -608,7 +608,7 @@ traversal:
 		if row := p.clients[cpath]; len(row) > 0 {
 			return row[0].client
 		}
-		c, pr := NewPromisedClient(pipelineClient{
+		c, pr := NewPromisedClient(PipelineClient{
 			p:         p,
 			transform: ft,
 		})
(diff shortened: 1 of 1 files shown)


answer.go
```diff
@@ -608,7 +608,7 @@ traversal:
 		if row := p.clients[cpath]; len(row) > 0 {
 			return row[0].client
 		}
-		c, pr := NewPromisedClient(pipelineClient{
+		c, pr := NewPromisedClient(PipelineClient{
 			p:         p,
```
lthibault, line 612: Are we intentionally keeping these fields private?  That is:  is it ever expected that users will instantiate a `PipelineClient` in their own code?

lthibault (approved):

answer.go
zenhack: I don't see a clear use case; we can always expose later. But I don't have a strong opinion.

lthibault: Yep, sounds good.  Just making sure it wasn't overlooked.  Resolving.

PR #212, opened 2022-01-02
Simplify how we determine whether to embargo

zenhack:
Rather than having recvCap deal with this directly, we now have a separate
routine that just checks if a client is local, which we invoke higher up
in the chain.

This means we're not trying to figure this out based on the cap
descriptor, which means (1) fewer cases, and (2), in the case of
a cap descriptor that denotes a promise or answer, we will be able
to recursively call this on the target -- whose original CapDescriptor
may not be known.

This introduces one minor functional change: ErrorClients are now
treated as local. This means we'll set up embargos for them, which
we don't *need* to do, but it shouldn't be a problem; calling methods
on errors is not a case that is terribly sensitive to the extra round
trip. We could try to rework ErrorClient to avoid this, but it doesn't
seem worth the trouble.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -1079,13 +1079,13 @@ func (c *Conn) handleFinish(ctx context.Context, id answerID, releaseResultCaps 
 // error indicates a protocol violation.
 //
 // The caller must be holding onto c.mu.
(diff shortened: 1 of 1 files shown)


lthibault (approved):
LGTM.  The readability improvement is significant.

PR #213, opened 2022-01-04
sendCap: emit receiverAnswer when appropriate.

zenhack:
I need to do some testing before we should merge this, but this patch should have us emitting `receiverAnswer` where appropriate. Marking as a draft now, but this is representative.


zenhack:
Added versions of the tests from before but that actually use the client interface, and the pass. I'm also experimenting with this in the application that motivated me to build all this out, and it works... intermittently. So there's still some debugging to do before this can be merged.

zenhack:
N.B. I've merged #215 into this; that pr should be reviewed & merged first

zenhack:
@lthibault, this is probably worth of review; see #216 wrt. the latest in the debugging saga.

lthibault (approved):

zenhack:
I've merged #219 into this, so it should be good to go, though that should be reviewed & merged on its own first.

lthibault (approved):

lthibault:
LGTM.  This is marked WIP, so holding off on the merge, just to be safe.

zenhack:
Ah, yeah, forgot to change that. I'll hit merge then.

PR #224, opened 2022-02-20
Bugfix/unbounded send queue

lthibault:
Fixes https://github.com/capnproto/go-capnproto2/issues/190 and https://github.com/capnproto/go-capnproto2/issues/194.


rpc/rpc.go
```diff
@@ -138,82 +132,104 @@ type ErrorReporter interface {
...
 			}
 		}
 	}()
+
 	return c
 }
 
+func (c *Conn) newTask(f func(context.Context) error) func() error {
+	return func() error {
+		defer c.tasks.Done()
```
zenhack, line 179: Where is the corresponding call to c.tasks.Add()? Maybe this should just be deleted now that newTask is called inside g.Go? But I don't think that's compatible with using startTask() elsewhere, so maybe not.

rpc/rpc.go
```diff
@@ -138,82 +132,104 @@ type ErrorReporter interface {
...
 	})
+
+	var err error
+	select {
+	case err = <-cherr:
+	case <-ctx.Done():
+		err = ctx.Err()
+	}
+
 	if err != nil {
```
zenhack, line 221: You need to be holding the lock here, to manipulate fields like c.questions.

rpc/rpc.go
```diff
@@ -354,113 +370,139 @@ closeTransport:
 	return nil
 }
 
+func (c *Conn) send(ctx context.Context) error {
+	for {
+		if err := c.sendMessage(ctx, func(msg rpccp.Message) error {
```
zenhack, line 375: Style nitpick: given the size of the expression, this would probably scan better as a separate assignment and then if statement, rather than using the compound if form.

zenhack (changes requested):
This looks like it's broadly on the right track. A couple things in-line.

exc/types.go
```diff
@@ -0,0 +1,83 @@
+package exc
+
+import (
+	"errors"
+	"strconv"
+)
+
+// Type indicates the type of error, mirroring those in rpc.capnp.
+type Type int
```
zenhack, line 9: I don't like re-declaring this by hand vs. just importing the generated code. Esp. with stuff like the String() method, which `rpc.Exception_Type` already has...

zenhack (commented):
Some immediate thoughts about the error handling package. Haven't looked through in detail yet, but will try to find time in the next couple days to do a more thorough review.

lthibault:
This is more or less ready for review (modulo the deadlock in `TestRecvAbort`).

This refactor ended up introducing an issue that prompted me to refactor capnp exceptions.  Briefly, `context.Canceled` is returned from send operations during shutdown, causing a number of tests to fail because they were expecting `nil` errors.  The fix was to wrap errors with `fmt.Errorf` and the `%w` verb instead of flattening everything to strings with `%v` or `err.Error()`.

Note that the use of the standard library `errors` package introduced a name collision with our own `internal/errors` package, which has already been an annoyance in the past.  More generally, I've historically found it quite difficult to inspect error values returned by the RPC code due to this "flattening" effect, and have been thinking quite a bit about how this could be improved.

Both of the above issues prompted me to make changes to `internal/errors`, with the goal of:

1.  Facilitating the inspection of error values using standard libarary `errors.Is` and `errors.As`.
2. Exporting an `Exception` type to users.
3. Eventually supporting a richer error system (e.g. providing a `SendError` type that wraps a more specific error type, itself a capnp Exception).

As a result, I performed the following actions:

1.  Renamed the `errors` package to `exc`  (short for "exception")
2. Renamed `exc.Error` to `Exception`
3. Implemented an `Annotator` type to facilitate package-level annotations (e.g. with the `"rpc"` prefix)
4. Refactored error handling to make use of the new `Annotator` type
5. Added an `exc.IsType` function that asserts exception types in a manner analogous to `errors.Is`.  Although I ended up not using it in this PR, I nevertheless opted to keep it in the code base as it will be useful for future improvements.

Let me know if any of this is unclear or seems harebrained!

exc/types.go
lthibault: Oh, I didn't realize the generated code already had this.  Is there a reason it wasn't used in the original `internal/errors` package?

zenhack: Ah, I didn't even notice you'd just moved it when making this comment. @zombiezen would have to weigh in wrt if there was a specific reason not to, but the only possible reason I see is that the names are a bit more verbose. But if nothing else I thing we should make these aliases...

Let's put this off until after this pr is merged though; it's already a bit off-topic. But since we're exporting the package, we should be intentional about this before we stabilize the API.

lthibault: It seems to introduce an import cycle too, so maybe that's the reason.

Just to be clear:  are you saying I should revert the changes to the error package or simply hold off on aliasing `Type` with `Exception_Type`? I'd prefer to at least keep the present changes, unless you strongly object.  They make it quite a bit easier to handle errors internally, which is quite helpful.

zenhack: Just hold off on aliasing.

rpc/rpc.go
```diff
@@ -137,83 +125,123 @@ type ErrorReporter interface {
...
+
+	// start background tasks
+	g.Go(c.backgroundTask(c.send))
+	g.Go(c.backgroundTask(c.receive))
+
+	// monitor background tasks
 	go func() {
-		abortErr := c.receive(c.bgctx)
-		c.tasks.Done()
+		err := g.Wait()
```
zenhack, line 158: Did you mean to swap the Done() for a Wait()? My hunch is this is the source of your deadlock.

lthibault: I think that's correct, but you've put me on the right track.  I think the issue is that the `errgroup.Group`'s context returns on the first non-nil error, so I need to handle the `context.Canceled` here instead of in `c.backgroundTask`.

Testing now.

lthibault: No dice.  The `Done()` you're looking for is in `backgroundTask`.  This just waits until either `c.send` or `c.recv` terminate.

lthibault:
Ok, I think this is now good to go.  🤞 

> I don't think that will ever happen? in particular, send doesn't return until the bgctx is canceled, which happens in shutdown, which is after called *after* this wait.

`bgctx` expires when either of the errgroup's goroutines returns a non-nil error.  There were actually two bugs here:

1. I mistakenly thought `bgctx` would expire on nil errors as well.
2. `bgctx` is _always_ expired when we enter `shutdown()`, so we need an additional flag to suppress duplicate calls.

Re bug 2:  I took the liberty of simplifying the `Close()` and `shutdown()` logic a bit, as well as apply some opinionated readability improvements.  In particular:

- Close no longer returns `ExcAlreadyClosed`.  Repeated calls to `Close()` are idempotent, and will simply return `nil`.
- The `c.closed` flag was renamed to `closing` and is used to detect duplicate calls to `shutdown`.

lthibault:
Ugh ... unit tests are still blocking with the `-race` flag.  Going to investigate a bit, but they definitely pass on my machine (`go1.17.7`).

lthibault:
Hmm, I'm also seeing quite a few data-races.  I'll investigate this later.  Input welcome, but I know you're busy so no rush.

zenhack:
Yeah, definitely need to fix the races. Looks like it's happening in Encoder.Encode, so my first guess is that you missed some spot outside of the send() goroutine that is still directly sending data on the transport. I will have more time to stare at it next week.

zenhack:
Ok, yeah, grepping for `c.transport.NewMessage` comes up with a bunch of places where we're sending a message outside of the send goroutine. So those will need to be fixed to delegate to the send goroutine.

rpc/rpc.go
```diff
@@ -1256,245 +1249,234 @@ func (c *Conn) handleRelease(ctx context.Context, id exportID, count uint32) err
...
+				return
+			}
+
+			iface := ptr.Interface()
+			if !iface.IsValid() || int64(iface.Capability()) >= int64(len(ans.resultCapTable)) {
+				err = rpcerr.Failedf("incoming disembargo: sender loopback requested on a capability that is not an import")
+				return
+			}
+
+			client := ans.resultCapTable[iface.Capability()].AddRef()
```
lthibault, line 1333: I spotted a `LEAK: leaked client created by AddRef on /Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/rpc.go:1333`, while messing around with benchmarks.  I ran the command `go test -bench=. ./...`, but the output doesn't make it clear exactly which test is triggering this.  I tried commenting out the `.AddRef()` part of the line and the error goes away without introducing any tests failures, but I'm not super confident.  Any thoughts?

Full output below:

```
 $ go test -bench=. ./...
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkTextMovementBetweenSegments-8              3892            317708 ns/op
BenchmarkMarshal-8                               2416864               499.6 ns/op           264 B/op          3 allocs/op
BenchmarkUnmarshal-8                             2150120               555.0 ns/op           192 B/op          3 allocs/op
BenchmarkUnmarshal_Reuse-8                       2972167               419.1 ns/op             0 B/op          0 allocs/op
BenchmarkDecode-8                                    340           3256078 ns/op         294.83 MB/s     2880312 B/op      40002 allocs/op
BenchmarkDecode_Reuse-8                              727           1662096 ns/op         577.58 MB/s         400 B/op          3 allocs/op
BenchmarkGrowth_SingleSegment-8                      164           7588045 ns/op         138.19 MB/s     4343117 B/op         10 allocs/op
BenchmarkGrowth_MultiSegment-8                       100          10035792 ns/op         104.49 MB/s     1648598 B/op         21 allocs/op
BenchmarkSmallMessage_SingleSegment-8            1735057               689.2 ns/op       104.47 MB/s        1352 B/op          5 allocs/op
BenchmarkSmallMessage_MultiSegment-8             2053214               583.9 ns/op       123.32 MB/s        1216 B/op          4 allocs/op
PASS
ok      capnproto.org/go/capnp/v3       16.372s
PASS
ok      capnproto.org/go/capnp/v3/capnpc-go     0.444s
PASS
ok      capnproto.org/go/capnp/v3/encoding/text 0.120s
PASS
ok      capnproto.org/go/capnp/v3/exc   0.130s
PASS
ok      capnproto.org/go/capnp/v3/flowcontrol   0.154s
?       capnproto.org/go/capnp/v3/internal/aircraftlib  [no test files]
?       capnproto.org/go/capnp/v3/internal/capnptool    [no test files]
?       capnproto.org/go/capnp/v3/internal/chanmutex    [no test files]
PASS
ok      capnproto.org/go/capnp/v3/internal/demo 0.138s
?       capnproto.org/go/capnp/v3/internal/demo/books   [no test files]
PASS
ok      capnproto.org/go/capnp/v3/internal/mpsc 0.231s
?       capnproto.org/go/capnp/v3/internal/nodemap      [no test files]
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3/internal/packed
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkPack-8                   102496             11517 ns/op         889.12 MB/s
BenchmarkUnpack-8                 127020              9691 ns/op        1056.62 MB/s
BenchmarkUnpack_Large-8           784611              1401 ns/op        37386.49 MB/s
BenchmarkReader-8                  66501             17918 ns/op         571.48 MB/s
BenchmarkReader_Large-8            35362             34034 ns/op        1538.70 MB/s
PASS
ok      capnproto.org/go/capnp/v3/internal/packed       8.488s
?       capnproto.org/go/capnp/v3/internal/schema       [no test files]
?       capnproto.org/go/capnp/v3/internal/strquote     [no test files]
?       capnproto.org/go/capnp/v3/internal/syncutil     [no test files]
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3/pogs
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkExtract-8         67496             17730 ns/op           15184 B/op         32 allocs/op
BenchmarkInsert-8          67862             17781 ns/op           15164 B/op         29 allocs/op
PASS
ok      capnproto.org/go/capnp/v3/pogs  2.991s
LEAK: leaked client created by AddRef on /Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/rpc.go:1333
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3/rpc
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkPingPong-8        44317             26844 ns/op            8888 B/op         96 allocs/op
PASS
exit status 1
FAIL    capnproto.org/go/capnp/v3/rpc   2.148s
?       capnproto.org/go/capnp/v3/rpc/internal/testcapnp        [no test files]
PASS
ok      capnproto.org/go/capnp/v3/schemas       0.117s
PASS
ok      capnproto.org/go/capnp/v3/server        0.123s
?       capnproto.org/go/capnp/v3/std/capnp/compat/json [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/cxx [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/persistent  [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/rpc [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/rpctwoparty [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/schema      [no test files]
?       capnproto.org/go/capnp/v3/std/capnp/stream      [no test files]
?       capnproto.org/go/capnp/v3/std/go        [no test files]
FAIL
```

zenhack: Hypothesis: this should be `client =` rather than `client :=`; this is a new variable that is shadowing the old one, and thus the call to Release() on the old variable ends up not doing what we want.

rpc/rpc.go
```diff
@@ -884,98 +918,60 @@ func (c *Conn) handleReturn(ctx context.Context, ret rpccp.Return, releaseRet ca
...
-			c.report(annotate(err, "incoming return"))
-			continue
-		}
-		err = send()
-		release()
-		if err != nil {
-			c.report(failedf("incoming return: send disembargo: %w", err))
-			continue
+	for _, s := range pr.disembargoes {
+		if err := c.sendMessage(ctx, s.buildDisembargo); err != nil {
```
zenhack, line 948: You're calling sendMessage from the receive loop here, rather than sendMessageAsync. Maybe move the error reporting into the callback?

rpc/rpc.go
```diff
@@ -884,98 +918,60 @@ func (c *Conn) handleReturn(ctx context.Context, ret rpccp.Return, releaseRet ca
...
-		err = send()
-		release()
-		if err != nil {
-			syncutil.With(&c.mu, func() {
-				c.unlockSender()
-				close(q.finishMsgSend)
-			})
-			c.report(failedf("incoming return: send finish: build message: %w", err))
-			return nil
+	err := c.sendMessage(ctx, func(m rpccp.Message) error {
```
zenhack, line 955: Same thing here; you're blocking the receive loop.

rpc/rpc.go
```diff
@@ -1256,245 +1249,234 @@ func (c *Conn) handleRelease(ctx context.Context, id exportID, count uint32) err
...
-			c.mu.Unlock()
-			client.Release()
-			return failedf("incoming disembargo: sender loopback requested on a capability that is not an import")
+			release()
+			return err
 		}
-		// TODO(maybe): check generation?
 
 		// Since this Cap'n Proto RPC implementation does not send imports
 		// unless they are fully dequeued, we can just immediately loop back.
```
zenhack, line 1355: You didn't add this but: I have no idea what this comment is trying to say, do you? (It sounds like it documents some important invariant, and I am therefore uncomfortable with the fact that I don't understand it).

rpc/rpc.go
```diff
@@ -1256,245 +1249,234 @@ func (c *Conn) handleRelease(ctx context.Context, id exportID, count uint32) err
...
 	}
+
+	return
 }
 
-// sendMessage creates a new message on the Sender, calls f to build it,
-// and sends it if f does not return an error.  When f returns, the
-// message must have a nil CapTable.  The caller must be holding onto
-// c.mu.  While f is being called, it will be holding onto the sender
-// lock, but not c.mu.
```
zenhack, line 1409: It would be good to retain some top level comment for this function, even if it needs adjusting. The roles and invariants around this and the async version are a bit subtle and worth documenting clearly.

rpc/rpc.go
```diff
@@ -1256,245 +1249,234 @@ func (c *Conn) handleRelease(ctx context.Context, id exportID, count uint32) err
...
-		case <-c.bgctx.Done():
-			c.mu.Lock()
-			return ExcClosed
-		}
-		c.mu.Lock()
+func (c *Conn) sendMessageAsync(ctx context.Context, f func(rpccp.Message) error) {
+	cherr, release := c.sender.Send(c.newSender(ctx, f))
+	defer release()
+
+	select {
```
zenhack, line 1431: Wait, am I misunderstanding? It looks like this *also* blocks, i.e. is not actually async. How is this supposed to work?

Up until reading this part I was assuming that this function was non-blocking, which was the whole point of this pr, but it looks like it just does the same thing as sendMessage, except it calls ReportError instead of returning the error.

rpc/rpc.go
```diff
@@ -1256,245 +1249,234 @@ func (c *Conn) handleRelease(ctx context.Context, id exportID, count uint32) err
...
 	releaseList(msg.CapTable).release()
 	msg.CapTable = nil
 }
+
+type sendQueue mpsc.Queue
+
+func (s *sendQueue) Send(sender func() error) (<-chan error, capnp.ReleaseFunc) {
+	cherr := make(chan error, 1)
+	s.Tx.Send(func() { cherr <- sender() })
+	return cherr, func() { /* TODO:  pool.Put(cherr) */ }
```
zenhack: I don't think it's worth complicating the API in anticipation of this.

zenhack (changes requested):
Comments in-line. A mix of minor style/docs things and also some major problems/questions.

zenhack (changes requested):
Some minor comments in-line (including bumps on a few things I'd asked about last time). Mostly looks good though.

rpc/rpc.go
```diff
@@ -137,84 +125,121 @@ type ErrorReporter interface {
...
-		c.tasks.Done()
+		err := g.Wait()
+
+		// Treat context.Canceled as a success indicator.
+		// Do not report or send an abort message.
+		if errors.Is(err, context.Canceled) {
+			err = nil
+		}
+
+		c.er.ReportError(err) // ignores errors
```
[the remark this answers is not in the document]
lthibault: Fixed.  Should have been "ignores nil errors".

rpc/bench_test.go
```diff
@@ -34,6 +34,7 @@ func BenchmarkPingPong(b *testing.B) {
 	client := testcp.PingPong{Client: conn2.Bootstrap(ctx)}
 	defer client.Client.Release()
 	b.ResetTimer()
+	b.ReportAllocs()
```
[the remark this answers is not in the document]
lthibault: Removed.

rpc/rpc.go
```diff
@@ -137,83 +125,124 @@ type ErrorReporter interface {
...
+	// while sending the bootstrap message.
 	if !c.startTask() {
-		c.mu.Unlock()
-		return capnp.ErrorClient(ExcClosed)
+		return capnp.ErrorClient(rpcerr.Disconnectedf("connection closed"))
 	}
 	defer c.tasks.Done()
+
 	q := c.newQuestion(capnp.Method{})
-	c.setAnswerQuestion(q.p.Answer(), q)
```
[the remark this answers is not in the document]
lthibault: Fixed.

Relatedly:  what does `setAnswerQuestion` achieve?  I'd like to add a docstring, as I'm struggling to remember the role of the metadata field.

rpc/answer.go
```diff
@@ -206,18 +202,17 @@ func (ans *answer) Return(e error) {
 // Finish with releaseResultCaps set to true, then sendReturn returns
 // the number of references to be subtracted from each export.
 //
-// The caller must be holding onto ans.c.mu and the sender lock.
-// The result's capability table must have been extracted into
-// ans.resultCapTable before calling sendReturn. Only one of
-// sendReturn or sendException should be called.
+// The caller MUST hold ans.c.mu.  The result's capability table MUST
+// have been extracted into ans.resultCapTable before calling sendReturn,
+// and call only one of sendReturn or sendException.
```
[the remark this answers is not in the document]
lthibault: >Only one of sendReturn or sendException should be called

This can be interpreted as a recommendation rather than the hard requirement that _either_ `sendReturn` or `sendException` _must_ be called.

How do you feel about the following?

>The caller MUST be holding onto ans.c.mu.  The result's capability table
MUST have been extracted into ans.resultCapTable before calling sendReturn.
sendReturn MUST NOT be called if sendException was previously called.

rpc/rpc.go
lthibault: Best guess:  this has to do with some kind of "version" of something?  Grepping through the C++ implementation reveals use of the term "generation" consistent with this meaning (see [here](https://github.com/capnproto/capnproto/search?q=generation)).  Notably, it seems to involve reconnection of some sort, so I suppose the "generation" refers to the number of prior connections, or something of the sort.

Does that ring any bells?

rpc/receiveranswer_test.go
```diff
@@ -110,6 +112,8 @@ func TestBootstrapReceiverAnswerRpc(t *testing.T) {
 }
 
 func TestCallReceiverAnswerRpc(t *testing.T) {
+	t.Parallel()
+
 	cClient, cServer := netPipe()
```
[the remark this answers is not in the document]
lthibault: >Trying it locally though, it still hangs...

Same on my end 😕

rpc/answer.go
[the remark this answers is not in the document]
zenhack: SGTM.

rpc/receiveranswer_test.go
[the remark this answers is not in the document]
zenhack: Should probably try this again.

rpc/rpc.go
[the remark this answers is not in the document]
zenhack: Grep for uses of getAnswerQuestion. Basically it's to help "unwrapping" capabilities so we can inspect them.

rpc/receiveranswer_test.go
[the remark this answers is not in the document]
lthibault: Yeah, I tried it right before posting my last message.  It still deadlocks with the asynchronous send.

zenhack (approved):

PR #225, opened 2022-02-20
Don't block when queuing methods on *Server.

### LINKED ISSUE #189: Non-deterministic deadlock on concurrent RPC calls

zenhack:
This should fix #189, though right now the tests are deadlocking and
I don't have the energy to debug, so I'm putting it down.

This also obviates MaxConcurrentCalls; in the current version of the
patch it is still present, but unused. Once this is working, a
subsequent commit will remove the Policy struct.

This significantly restructures the internals of *Server, and I
think the result is easier to understand: the methods on *Server
that start calls just stick the calls in a queue, and there is a
goroutine that pulls them out and handles them. This even gets rid
of Server.mu entirely, since nothing actually needed the lock
anymore!

Hopefully it will stay simple when I actually get it working.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -85,9 +85,12 @@ type Server struct {
 	shutdown Shutdowner
 
 	// Cancels handleCallsCtx
-	cancelCalls context.CancelFunc
+	cancelHandleCalls context.CancelFunc
 
-	// Context used by the goroutine running handleCalls()
+	// Context used by the goroutine running handleCalls(). Note
+	// the calls themselves will have different contexts, which
+	// are not children of this context, but are supplied by
+	// start(). See cancelCurrentCall.
 	handleCallsCtx context.Context
 
 	// wg is incremented each time a method is queued, and
@@ -97,6 +100,15 @@ type Server struct {
 	// Calls are inserted into this queue, to be handled
 	// by a goroutine running handleCalls()
 	callQueue *mpsc.Queue[*Call]
+
+	// When a call is in progress, this channel will contain the
+	// CancelFunc for that call's context. A goroutine may receive
+	// on this to fetch the function, and is then responsible for calling
+	// it. This happens in Shutdown().
+	//
+	// The caller must call cancelHandleCalls() *before* calling
+	// the received CancelFunc.
+	cancelCurrentCall chan context.CancelFunc
 }
 
 // Policy is a set of behavioral parameters for a Server.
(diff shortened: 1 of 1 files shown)


zenhack:
Got myself a stack trace, and several of the goroutines are blocked in `Conn.lockSender()` -- so I think I'm going to put this aside until #224 is merged, since it removes that function entirely and might fix things.

zenhack:
Rebased, but unfortunately it's still deadlocking. The test that is deadlocking is `server.TestServerShutdown`. I don't really have time to debug it right now though, so will have to get back to it (maybe tomorrow).

zenhack:
Alright, fixed it. The fix is fiddlier than I would like; see the comments about ordering in the last commit. Open to suggestions for simpler solutions.

I have let `Policy` live for now, mainly to keep the diff small; a followup pr will do the requisite slashing and burning.

lthibault (approved):
LGTM.

It took me a while to grok the exact behavior of `handleCall(s)` and `Ack()`, so I'm wondering if it might not be a good idea to document the high-level process somewhere?  Perhaps as a docstring for `handleCalls`?

Below is my understanding of the process, which I want to sanity-check with you, and which might also serve as a starting point for such a docstring.

1. The default behavior is to process incoming method calls synchronously.
2. `c.Ack()` signals that the server start processing the next queued method call, and that `c` can continue concurrently in the background.
3. `Ack()` spawns another call to `handleCalls` in a separate goroutine
4. `handleCalls` blocks on `srv.handleCall`, until the call completes; if it determines that ack was called, it returns since another instance of `handleCalls` is running in the background (via step 3).

zenhack:
Quoting Louis Thibault (2022-04-18 12:52:14)

>    Below is my understanding of the process, which I want to sanity-check
>    with you, and which might also serve as a starting point for such a
>    docstring.
>     1. The default behavior is to process incoming method calls
>        synchronously.
>     2. c.Ack() signals that the server start processing the next queued
>        method call, and that c can continue concurrently in the
>        background.
>     3. Ack() spawns another call to handleCalls in a separate goroutine
>     4. handleCalls blocks on srv.handleCall, until the call completes; if
>        it determines that ack was called, it returns since another
>        instance of handleCalls is running in the background (via step 3).

Correct. And yeah, it could use high-level docs; I'll do that soonish.

lthibault:
> Quoting Louis Thibault (2022-04-18 12:52:14)
> Below is my understanding of the process, which I want to sanity-check with you, and which might also serve as a starting point for such a docstring. 1. The default behavior is to process incoming method calls synchronously. 2. c.Ack() signals that the server start processing the next queued method call, and that c can continue concurrently in the background. 3. Ack() spawns another call to handleCalls in a separate goroutine 4. handleCalls blocks on srv.handleCall, until the call completes; if it determines that ack was called, it returns since another instance of handleCalls is running in the background (via step 3).
> Correct. And yeah, it could use high-level docs; I'll do that soonish.

Ok perfect. Merging in the meantime, as this is just a detail anyway.

PR #226, opened 2022-03-16
Use generics for mpsc

zenhack:
This also updates our language version to 1.18 and changes what versions
we run against in the build matrix.

---

Going forward, we'll want to use this in generated code with capnp generics, but this seemed like a nice excuse to take Go's generics for a spin.


lthibault (approved):
I love it 👍 🎉

PR #228, opened 2022-03-26
Use the embed package to do less in mktemplates.

zenhack:
...with a bit more work we can probably get rid of the mktemplates
command entirely, but this was low hanging fruit.

Note that this also re-generates some of the .capnp.go files. The
differences are I assume because I am using a new version of the capnp
compiler than when these were last generated? It seems nothing changed
except for the exact bytes of the schema; the actual generated *code*
is the same. We really should have CI make sure these are correct somehow.


zenhack:
Update, went ahead and finished the job; mktemplates is gone.

lthibault:
This looks promising 🙂   I'm going to be traveling next week, and will have limited availability, but I'll give this a thorough review when I get back.

lthibault:
We should probably change [this import](https://github.com/capnproto/go-capnproto2/pull/228/files#diff-f505c679c0875544b7a396ddf3a3907ac82d6b2586c5bca44ab91176caa6dcb4R37) to `"context"`.

lthibault (changes requested):
LGTM overall.  See inline comment regarding context package.

zenhack:
I am confused; I don't see an import statement at that link?

lthibault:
> I am confused; I don't see an import statement at that link?

Oh, strange.  It's line 36:  `contextImport = "golang.org/x/net/context"`.

zenhack:
Rebased on top of master to fix the merge conflicts. That should also pull in the change in #227, which deals with that import.

lthibault (approved):

PR #229, opened 2022-04-04
Use generics for list types.

zenhack:
This defines two generic types in the main package, `EnumList` and `StructList`, and then makes the generated `Foo_List` types aliases for one or the other, depending on the representation of the type.

I'd like to add the ability to do `NewStructList[Foo](seg, length)`, but that requires NewStructList to be able to get the correct sizes for the struct `Foo`; doable, but I'm going to punt to a later patch.


lthibault (approved):
LGTM.

PR #232, opened 2022-04-28
Represent :Capability type as *capnp.Client.

lthibault:
Addresses https://github.com/capnproto/go-capnproto2/issues/192.

### DIFF
--- a/capnpc-go/capnpc-go.go
+++ b/capnpc-go/capnpc-go.go
@@ -15,7 +15,6 @@ import (
 	"flag"
 	"fmt"
 	"go/format"
-	"io"
 	"math"
 	"os"
 	"path/filepath"
@@ -372,15 +371,30 @@ func (g *generator) Value(rel *node, t schema.Type, v schema.Value) (string, err
 		var buf bytes.Buffer
 
 		sd, err := g.data.copyData(data)
-		if err == nil {
-			switch t.AnyPointer().Which() {
-			case schema.Type_anyPointer_Which_unconstrained:
-				err = g.anyPointerUnconstrained(&buf, t, sd)
+		if err != nil {
+			return "", err
+		}
 
-			default:
-				err = g.pointerValue(&buf, sd)
+		switch t.AnyPointer().Which() {
+		// Unconstrained?
+		case schema.Type_anyPointer_Which_unconstrained:
+
+			switch t.AnyPointer().Unconstrained().Which() {
+			// Capability type?
+			case schema.Type_anyPointer_unconstrained_Which_capability:
+				err = templates.ExecuteTemplate(&buf, "capability", capabilityParams{
+					G:     g,
+					Value: sd,
+				})
+				return buf.String(), err
 			}
 		}
+
+		// Fall back to default case => generic pointer value
+		err = templates.ExecuteTemplate(&buf, "pointerValue", pointerValueParams{
+			G:     g,
+			Value: sd,
+		})
 
 		return buf.String(), err
 
@@ -404,26 +418,6 @@ func (g *generator) Value(rel *node, t schema.Type, v schema.Value) (string, err
 	default:
 		return "", fmt.Errorf("unhandled value type %v", t.Which())
 	}
-}
-
-func (g *generator) anyPointerUnconstrained(w io.Writer, t schema.Type, sd staticDataRef) error {
-	switch t.AnyPointer().Unconstrained().Which() {
-	case schema.Type_anyPointer_unconstrained_Which_capability:
-		return templates.ExecuteTemplate(w, "capability", capabilityParams{
-			G:     g,
-			Value: sd,
-		})
-
-	default:
-		return g.pointerValue(w, sd)
-	}
-}
-
-func (g *generator) pointerValue(w io.Writer, sd staticDataRef) error {
-	return templates.ExecuteTemplate(w, "pointerValue", pointerValueParams{
-		G:     g,
-		Value: sd,
-	})
 }
 
 func (g *generator) defineAnnotation(n *node) error {
@@ -615,6 +609,19 @@ func (g *generator) defineField(n *node, f field) (err error) {
 				return err
 			}
 		}
+
+		switch t.AnyPointer().Which() {
+		// Unconstrained?
+		case schema.Type_anyPointer_Which_unconstrained:
+
+			switch t.AnyPointer().Unconstrained().Which() {
+			// Capability type?
+			case schema.Type_anyPointer_unconstrained_Which_capability:
+				return g.r.Render(structInterfaceFieldParams(params))
+			}
+		}
+
+		// Fall back to default case => generic pointer value
 		return g.r.Render(structPointerFieldParams{
 			structFieldParams: params,
 			Default:           defref,
--- a/capnpc-go/capnpc-go.go
+++ b/capnpc-go/capnpc-go.go
@@ -382,11 +382,7 @@ func (g *generator) Value(rel *node, t schema.Type, v schema.Value) (string, err
 			switch t.AnyPointer().Unconstrained().Which() {
 			// Capability type?
 			case schema.Type_anyPointer_unconstrained_Which_capability:
-				err = templates.ExecuteTemplate(&buf, "capability", capabilityParams{
-					G:     g,
-					Value: sd,
-				})
-				return buf.String(), err
+				return "nil", nil
 			}
 		}
 
@@ -617,7 +613,8 @@ func (g *generator) defineField(n *node, f field) (err error) {
 			switch t.AnyPointer().Unconstrained().Which() {
 			// Capability type?
 			case schema.Type_anyPointer_unconstrained_Which_capability:
-				return g.r.Render(structInterfaceFieldParams(params))
+				// params.FieldType = "*capnp.Client"
+				return g.r.Render(structCapabilityFieldParams(params))
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


lthibault:
No idea why tests are failing.  They pass on my machine 😕

capnpc-go/capnpc-go.go
```diff
@@ -372,16 +371,27 @@ func (g *generator) Value(rel *node, t schema.Type, v schema.Value) (string, err
...
+		switch t.AnyPointer().Which() {
+		// Unconstrained?
+		case schema.Type_anyPointer_Which_unconstrained:
 
-			default:
-				err = g.pointerValue(&buf, sd)
+			switch t.AnyPointer().Unconstrained().Which() {
+			// Capability type?
+			case schema.Type_anyPointer_unconstrained_Which_capability:
+				return "nil", nil
```
zenhack, line 385: This could probably use a comment explaining why it's ok to assume the value is nil.

capnpc-go/capnpc-go.go
```diff
@@ -615,6 +605,20 @@ func (g *generator) defineField(n *node, f field) (err error) {
 				return err
 			}
 		}
+
+		switch t.AnyPointer().Which() {
+		// Unconstrained?
```
zenhack, line 610: Since there's only one `case` here, this should probably just be an if statement. Which I think would also obviate this comment, which feels excessive to me. Same for the nested switch, and elsewhere where this pattern gets used.

capnpc-go/capnpc-go.go
```diff
@@ -751,6 +754,18 @@ func makeTypeRef(t schema.Type, rel *node, nodes nodeMap) (typeRef, error) {
 		case schema.Type_Which_anyPointer, schema.Type_Which_list, schema.Type_Which_interface:
 			return typeRef{name: "PointerList", newfunc: "NewPointerList", imp: capnpImportSpec}, nil
 		}
+	case schema.Type_Which_anyPointer:
+		switch t.AnyPointer().Which() {
+		case schema.Type_anyPointer_Which_unconstrained:
+
+			switch t.AnyPointer().Unconstrained().Which() {
+			case schema.Type_anyPointer_unconstrained_Which_capability:
+				return typeRef{name: "Client", imp: capnpImportSpec}, nil
```
zenhack, line 763: won't this return `Client`? I think we want `*Client`.

Edit: Ah, I see, you worked around this by putting the `*` in the template. It feels like a bit of a hack...

internal/aircraftlib/aircraft.capnp.go
```diff
@@ -2245,11 +2245,12 @@ func (s Z) SetAnyList(v capnp.Ptr) error {
 	return s.Struct.SetPtr(0, v)
 }
 
-func (s Z) AnyCapability() (capnp.Ptr, error) {
+func (s Z) AnyCapability() *capnp.Client {
```
zenhack, line 2248: I like that this patch removes a case where you have to check an error.

pogs/pogs_test.go
```diff
@@ -1227,11 +1227,8 @@ func zfill(c air.Z, g *Z) error {
 	case air.Z_Which_anyList:
 		return c.SetAnyList(g.AnyList.ToPtr())
 	case air.Z_Which_anyCapability:
-		if g.AnyCapability == nil {
-			return c.SetAnyCapability(capnp.Ptr{})
-		}
 		cap := capnp.NewInterface(c.Segment(), c.Message().AddCap(g.AnyCapability))
-		return c.SetAnyCapability(cap.ToPtr())
+		return c.SetAnyCapability(cap.Client())
```
zenhack, line 1231: I think this whole `case` could just be simplified to `c.SetAnyCapability(g.AnyCapability)`. This is the same logic that's moved into the setter.

rpc/receiveranswer_test.go
```diff
@@ -81,7 +79,7 @@ func TestBootstrapReceiverAnswerRpc(t *testing.T) {
 
 	res, rel := c.Call(ctx, func(p testcapnp.CapArgsTest_call_Params) error {
 		capId := p.Message().AddCap(c.Client.AddRef())
-		p.SetCap(capnp.NewInterface(p.Segment(), capId).ToPtr())
+		p.SetCap(capnp.NewInterface(p.Segment(), capId).Client())
```
zenhack, line 82: Again, I think you can skip the AddCap/NewInterface dance and just do `SetCap(c.Client.AddRef())`.

rpc/receiveranswer_test.go
```diff
@@ -126,7 +124,7 @@ func TestCallReceiverAnswerRpc(t *testing.T) {
 	self := selfRes.Self()
 	callRes, rel := self.Call(ctx, func(p testcapnp.CapArgsTest_call_Params) error {
 		capId := p.Message().AddCap(self.Client.AddRef())
-		p.SetCap(capnp.NewInterface(p.Segment(), capId).ToPtr())
+		p.SetCap(capnp.NewInterface(p.Segment(), capId).Client())
```
zenhack, line 127: Same thing here.

zenhack (changes requested):
Some minor things in line, but otherwise LGTM; thanks for doing this. Obviously we also need to figure out why the tests are failing.

capnpc-go/capnpc-go.go
lthibault: Yeah, it's not _great_.  I looked for a means of specifying a pointer in `capnpImportSpec`, but didn't see anything obvious.  Notably, `importSpec{path: capnpImport, name: "*capnp"}` produces an error.

I'm resolving this, but I'm happy to make improvements if you have any suggestions.

capnpc-go/capnpc-go.go
lthibault: My intent was to progressively add the other cases over the next few weeks.  I suppose I can always revert to a `switch` statement at that time, though.

capnpc-go/capnpc-go.go
lthibault: Fixed.

rpc/receiveranswer_test.go
lthibault: Fixed.

rpc/receiveranswer_test.go
lthibault: Fixed.

internal/aircraftlib/aircraft.capnp.go
lthibault: Yeah, same!  In fact, I was just wondering if we could similarly ignore errors in `SetAnyCapability`?  I'm guessing _not_, since pointers require jumping to a different offset in the segment, which may or may not exist.  Is that right?  Or is there another reason as well?

lthibault:
> Some minor things in line, but otherwise LGTM; thanks for doing this.

My pleasure, this was a fun one!

>Obviously we also need to figure out why the tests are failing.

I'll take a look at that now.  I'm admittedly quite stumped, though, as I can't reproduce the failure on my machine.

lthibault:
AFAICT the tests are actually _passing_, but the CI script is (sometimes?) interpreting calls to `t.Logf` as errors.

Notice the last line 👇 
```
=== RUN   TestMessage_ResetReadLimit
Error:     readlimit_test.go:70:    m := &Message{TraverseLimit: 42}
Error:     readlimit_test.go:72:    m.canRead(42) -> true
Error:     readlimit_test.go:74:    m.ResetReadLimit(8)
Error:     readlimit_test.go:76:    m.canRead(8) -> true
    readlimit_test.go:81: 
Error:     readlimit_test.go:84:    m := &Message{TraverseLimit: 42}
Error:     readlimit_test.go:86:    m.canRead(40) -> true
Error:     readlimit_test.go:88:    m.ResetReadLimit(8)
Error:     readlimit_test.go:92:    m.canRead(9) -> false
    readlimit_test.go:95: 
Error:     readlimit_test.go:98:    m := &Message{TraverseLimit: 42}
Error:     readlimit_test.go:100:    m.canRead(42) -> true
Error:     readlimit_test.go:102:    m.ResetReadLimit(8)
Error:     readlimit_test.go:104:    m.canRead(8) -> true
    readlimit_test.go:109: 
Error:     readlimit_test.go:112:    m := &Message{TraverseLimit: 42}
Error:     readlimit_test.go:114:    m.canRead(40) -> true
Error:     readlimit_test.go:116:    m.ResetReadLimit(8)
Error:     readlimit_test.go:120:    m.canRead(9) -> false
    readlimit_test.go:123: 
Error:     readlimit_test.go:126:    m := new(Message)
Error:     readlimit_test.go:128:    m.ResetReadLimit(0)
Error:     readlimit_test.go:132:    m.canRead(0) -> true
    readlimit_test.go:135: 
Error:     readlimit_test.go:138:    m := new(Message)
Error:     readlimit_test.go:140:    m.ResetReadLimit(0)
Error:     readlimit_test.go:144:    m.canRead(1) -> false
--- PASS: TestMessage_ResetReadLimit (0.00s)
```

capnpc-go/capnpc-go.go
zenhack: I'm fine leaving this as-is though I don't love it.

internal/aircraftlib/aircraft.capnp.go
zenhack: I think so, and in fact we could do the same for all interface types. Other pointer types could fail because they might need to allocate a landing pad for a far pointer, and the allocation could fail, but cap pointers don't ever need to do this so this can't actually fail.

zenhack:
Hm, looking at this I don't think the t.Log is the issue: see https://github.com/capnproto/go-capnproto2/runs/6230315382?check_suite_focus=true#step:4:1616; it looks like the race detector is what's complaining.

Otherwise LGTM, but we need to figure out what's going on there.

lthibault:
> https://github.com/capnproto/go-capnproto2/runs/6230315382?check_suite_focus=true#step:4:1616

Oh weird, I must have scrolled right over that.  Why isn't that happening on my end, I wonder ... 🤔 

Digging into this now.

zenhack (approved):

PR #234, opened 2022-05-03
Fix compilation error and test on intel 32-bit

Matherunner:
This PR fixes the following.

Set

```bash
export GOARCH=386
```

Then run `go test ./...`, it will fail:

```
./message.go:790:14: cannot convert maxSegmentSize (constant 4294967288 of type Size) to type int
```

After fixing the compilation error, this test will fail:

```
--- FAIL: TestNextAlloc (0.00s)
    message_test.go:909: max size: nextAlloc(0, 9223372036854775800, 4294967288) = _, capnp: alloc 4294967288 bytes: too large; want >=4294967288, <nil>
```

Anecdotally, the fixed version on a 32-bit Windows application runs fine when encoding, decoding, and doing basic RPC.


zenhack (approved):
LGTM. Bonus points if you want to add 386 to our CI, so it catches word size problems in the future.

Matherunner:
> LGTM. Bonus points if you want to add 386 to our CI, so it catches word size problems in the future.

Good idea. Added 386 to the matrix and tested on a [PR](https://github.com/Matherunner/go-capnproto2/pull/1) on my fork.

lthibault:
@zenhack CI seems to be failing on the race condition we discovered a few days ago.  I'm OK to merge this, since it's definitely not causing any new issues.

zenhack (approved):

zenhack:
Yeah, I agree; that's an existing failure and this doesn't touch anything but the tests so it's pretty low risk. Merging. Thanks @Matherunner!

PR #238, opened 2022-05-08
flowcontrol.fixedLimiter: Swap in semaphore.Weighted.

zenhack:
...sadly, this does not fix my deadlock, but even so, no need to re-implement this ourselves.


zenhack:
That test doesn't use fixedLimiter, so I'm pretty confident. Merging.

PR #239, opened 2022-05-12
Fix race condition in handleReturn

lthibault:


lthibault:
@zenhack I have a lingering getting ref leak in `TestSendBootstrapCallException`:

```
LEAK: leaked client created by AddRef on /Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/export.go:160
```

Any thoughts?

rpc/rpc.go
```diff
@@ -591,7 +591,7 @@ func (c *Conn) handleBootstrap(ctx context.Context, id answerID) error {
 		syncutil.Without(&c.mu, rl.release)
 		return nil
 	}
-	if err := ans.setBootstrap(c.bootstrap.AddRef()); err != nil {
+	if err := ans.setBootstrap(c.bootstrap); err != nil {
```
zenhack, line 594: I'm still not convinced this is correct; my intuition is that this *should* be needed, since we want the caller to get their own reference, rather than stealing ours.

It makes sense that this fixes it given that you removed a Release() elsewhere, but I want to make sure there's not just another failure to call .Release() somewhere that's cancelling this out.

If this is correct I think it's important for us to be able to articulate why.

rpc/rpc.go
```diff
@@ -1092,22 +1092,25 @@ func (c *Conn) handleFinish(ctx context.Context, id answerID, releaseResultCaps
...
-	if ans.releaseMsg != nil {
-		ans.releaseMsg()
+	delete(c.answers, ans.id)
+
+	if ans.flags&releaseResultCapsFlag != 0 {
+		for id, count := range ans.exportRefs {
+			if _, e := c.releaseExport(id, count); err == nil && e != nil {
+				err = rpcerr.Annotate(err, "incoming finish: release result caps")
+			}
+		}
```
zenhack, line 1106: Looks like you inlined the body of ans.destroy() and then tweaked?

...I feel like it's worth auditing the other calls to .destroy() to make sure they're not doing the same wrong thing with the returned releaseList -- and maybe just change destory() and keep using it here?

zenhack (changes requested):
A couple comments.

lthibault:
@zenhack I've lost track of this issue.  Is it still relevant, or has this been resolved elsewhere?

PR #240, opened 2022-05-13
Refactor rpc.Conn lifecycle management

lthibault:
As per discussion on Matrix, this PR adds a `process` type that manages `rpc.Conn`'s lifecycle and teardown logic.  It is responsible for:

- Tasks
- Teardown logic
- Synchronization of connection shutdown

@zenhack This is extremely WIP, and currently breaks just about everything in the `rpc` package.  I'd appreciate some early feedback on the design, and if you approve of the overall direction of this PR, I'll keep working on it.  If that's the case, feel free to point out any obvious bugs you encounter.  Let me know what you think!


lthibault:
Correction:  I changed one line and now everything works again, except `TestSendAbort/ReceiverListening`.  How I managed to pull this off is beyond me -- must be that 2:00 AM programming magic 😄

lthibault:
Jotting down my thoughts to that I don't forget:

There is technically a data race with `process.AddRef`

```go
func (p *process) AddRef() bool {
	select {
	case <-p.root.Closing():
		return false
	default:
		p.refs.Add(1)
		return true
	}
}
```

Since there's no locking, it's possible for `process.signal(c)` (see `newProcess`) to happen (1) concurrently with or (2) before `p.refs.Add(1)`, at which point we would correspondingly observe:

1. Panic due to concurrent `sync.WaitGroup.Wait()` / `sync.WaitGroup.Done()` calls
2. The reference acquisition would succeed, but the process would not wait for the reference to be released before terminating.

Synchronizing this is a bit tricky because there are both:
1. ordering invariants across the cleanup logic (e.g. `refs.Wait()` must happen before `drainQueue`)
2. locking invariants across cleanup logic (e.g. `refs.Wait()` must be _unlocked_ and `drainQueue` must be _locked_)

To be clear:  this complexity always existed, and I think it's a net win that it is now grouped in a single place.

In order to avoid the data-race, we need to make sure that no calls to `AddRef` can take place after the root `p.root` has begun to shut down, _i.e._ after `<-p.Closing()` unblocks.  The challenge is that this happens only after `p.root`'s process function has returned.  In particular:  `p.signal` is not guaranteed to return before `<-p.Closing()` unblocks.

To avoid data races, I _think_ it is sufficient to require callers to `AddRef` and `withRef` to hold `c.mu`.

lthibault:
```
ok      capnproto.org/go/capnp/v3/pogs  (cached)
--- FAIL: TestSendAbort (0.00s)
    --- FAIL: TestSendAbort/ReceiverListening (0.00s)
        level0_test.go:78: recvMessage(ctx, p2): rpc pipe: receive on closed pipe
```

Tests are now passing except for this.  Insights are most welcome.

lthibault:
Superseded by #251.

PR #242, opened 2022-05-13
Export Client.h.done channel through Done() method.

lthibault:
This commit adds a `Client.Done() <-chan struct{}`  method that returns `c.h.done`.  This allows application code to perform cleanup when clients are released.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -379,7 +379,8 @@ func (c *Client) IsSame(c2 *Client) bool {
 	return h1 == h2
 }
 
-// Done returns when the client has been shut down.
+// Done returns when the client begins shutting down.  The channel
+// returned by Done() is closed before ClientHook.Shutdown() is called.
 func (c *Client) Done() <-chan struct{} { return c.h.done }
 
 // Resolve blocks until the capability is fully resolved or the Context is Done.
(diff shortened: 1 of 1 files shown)


capability.go
```diff
@@ -379,6 +379,10 @@ func (c *Client) IsSame(c2 *Client) bool {
 	return h1 == h2
 }
 
+// Done returns when the client begins shutting down.  The channel
+// returned by Done() is closed before ClientHook.Shutdown() is called.
```
zenhack, line 383: If we're going to expose this I think we need to explain the lifecycle semantics a bit better. Also skimming through it seems like this might interact with promise resolution in some ways that are not terribly obvious? Anyway it's a bad sign that I had to go dig through capability.go to work out what this actually does.

zenhack (commented):
Do you have a particular use case in mind for this?

zenhack:
One more thing: basically every other method on `Client` works fine if the receiver is nil. I think we could/should do the same here just by having a global variable that's pre-closed that we return in that case.

lthibault:
As discussed on Matrix, this ended up not being needed, so I'm closing this PR.

However, I think this is a feature to keep in mind, as the only way to detect a released `Client` is to try calling some method.  Thus, there's no way to receive an asynchronous notification.

PR #247, opened 2022-05-21
Add typed wrapper for lists of capabilities.

zenhack:
...To match the ones for structs and enums.

---

We should add a similar wrapper for nested lists (`List (List(T)`) before we release v3, and any other such missing list types I haven't thought of.


lthibault:
Would it make sense to have a `List[Client]` as well, now that we correctly handle `:Capability` types?

lthibault (approved):

PR #250, opened 2022-05-24
Remove redundant call to setAnswerQuestion

zenhack:
There's already a call to this in newQuestion, so this is unnecessary.

---

@lthibault, I seem to recall you having submitted a PR that removed this line, and that change got backed out. Staring at it with nothing else going on, it's clear that this change makes sense, I guess I didn't process it at the time since the change seemed unrelated.


zenhack:
Hm, test failures. Going to mark as a draft until I can investigate further.

lthibault:
I wonder if these tests might not be fixed by #261.

zenhack:
We can check once it's merged.

lthibault:
Just tested against `main` and it seems like #261 did the trick! 🎉 
Feel free to rebase & merge when you're ready.  I don't think it requires another review.

zenhack:
Alright, I merged main back in and if CI passes I'll hit the button.

PR #251, opened 2022-05-24
Experiment with structured process lifecycle management.

zenhack:
@lthibault, I started playing with my own ideas about how to clean up process lifecycle management. I'm curious to your thoughts, particularly regarding.

- Whether you think the proc package is The Right Abstraction, and whether it will actually help with the rest of the shutdown logic.
- If not, what does `goprocess` give us that this doesn't?

Part of this is my attempt to understand the problem space.

---

This patch introduces a `proc` package under internal, with an
abstraction I think will be useful in cleaning up the shutdown logic.

It also includes a rework of the send goroutine (in sender.go) that
reflects how I *want* this to work (notably, the logic for drainQueue
runs in the send gorotuine itself), and how the proc package can be used
to help with that. The new sender is dead code; I intend this as a
starting point for discussion.

### DIFF
--- a/rpc/internal/proc/proc.go
+++ b/rpc/internal/proc/proc.go
@@ -1,0 +1,103 @@
+// Package proc assits managing the lifecycle of goroutines.
+package proc
+
+import (
+	"context"
+	"sync"
+)
+
+// Spawn a new process.
+//
+// A process can be in one of the following states:
+//
+// - Running
+// - Stopping
+// - Stopped
+//
+// ...starting in the Running state.
+//
+// f is a callback that implements the actual logic of the process. It will be
+// passed a child context of ctx, and a Self. It should:
+//
+// 1. Run until its context is canceled is canceled (or its work is done).
+// 2. invoke BeginShutdown() on the Self that was passed to it, which
+//    transitions the process to the Stopping state.
+// 3. Perform any shutdown logic that is needed.
+// 4. Return. This transitions the process into the Stopped state.
+//
+// Returns a handle, which can be used by clients of the process to interact
+// with its lifecycle.
+func Spawn(ctx context.Context, f func(context.Context, Self)) Handle {
+	ctx, cancel := context.WithCancel(ctx)
+	proc := &proc{
+		cancel: cancel,
+		done:   make(chan struct{}),
+	}
+	handle := Handle{proc}
+	self := Self{proc}
+	go func() {
+		defer func() {
+			self.BeginShutdown()
+			close(proc.done)
+		}()
+		f(ctx, self)
+	}()
+	return handle
+}
+
+// A Handle is used for interacting with a process.
+type Handle struct {
+	proc *proc
+}
+
+// Cancel cancels the process's context.
+func (h Handle) Cancel() {
+	h.proc.cancel()
+}
+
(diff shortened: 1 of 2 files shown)


zenhack:
Note that this is *complete* untested (though it compiles), and should not be merged; it's really for discussion only.

zenhack:
Pushed a tweak; this now also protects some process-specific state, which transfers ownership during shutdown. I think this will be really nice to work with.

zenhack:
...I wonder if we shouldn't be agressively pulling stuff out of the rpc package and into small, self contained packages under `internal/`; sender.go is pretty self contained, and it would be nice to have clearer abstraction boundaries within the rpc subsystem.

lthibault:
I haven't looked at this yet, but I can respond to a couple of points:

>Whether you think the proc package is The Right Abstraction, and whether it will actually help with the rest of the shutdown logic.  If not, what does goprocess give us that this doesn't?

There's two things I like about `goprocess`.

1. It provides a clean abstraction for distinguishing between "I am in the process of shutting down" and "I have completely shut down".
2. It provides a clear ordering of goroutine states in a process tree.  For example, `<-parent.Closing()` happens before `<-child.Closing()`.
3. It provides a clear ordering of teardown logic in a process tree.  Namely:  `child.TeardownFunc()` happens before `parent.TeardownFunc()`

I don't have strong feelings about using this library in particular, and I'm very much open to reimplementing some or all of its internally.

>...I wonder if we shouldn't be agressively pulling stuff out of the rpc package and into small, self contained packages under internal/; sender.go is pretty self contained, and it would be nice to have clearer abstraction boundaries within the rpc subsystem.

I am very much in favor of this.

rpc/internal/proc/proc.go
```diff
@@ -0,0 +1,115 @@
...
+		cancel: cancel,
+		done:   make(chan struct{}),
+		state:  state,
+	}
+	handle := Handle[S]{proc}
+	self := Self[S]{proc}
+	go func() {
+		defer func() {
+			self.BeginShutdown()
+			close(proc.done)
```
lthibault, line 46: I think these should probably be two separate `defer` statements, because we want to make sure `close(proc.done)` happens despite any panics in `f()`.

lthibault (commented):
I really like this.  I flagged a minor issue inline, but overall, I think this is the right direction.

A follow up thought with respect to my previous comment:

Another thing I really like about `goprocess` is that process trees make it quite natural to express tasks:  you simply spawn a child off of the root process; this becomes a nop when the root is in the `Stopping` state.

Seems like this is covered by `WithLive`.

rpc/internal/proc/proc.go
zenhack: Done.

zenhack:
I should probably test it still, at least.

lthibault:
@zenhack Is this still planned?

zenhack:
I still think it's a good idea, it just hasn't made it to the top of my priorities list. Are you blocked on this for something you have in mind, or are you just trying to triage open prs?

zenhack:
Ok. No promises on timeline, but it's not actually api visible anyway (as of this patch, it's actually just dead code).

PR #254, opened 2022-06-14
Fix broken implemenation of .Ack()

zenhack:
I apparently completely broke this when I reworked the server internals to use an unbounded queue, as a result of some convoluted shutdown logic.

The first commit adds a test, verifying that calling Ack() does indeed unblock future calls. The test deadlocks.

The second commit  fixes the test, by replacing said convoluted shutdown logic with a more straightforward implementation that works.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -23,7 +23,6 @@ type Method struct {
 // A Call cannot be used after the server method returns.
 type Call struct {
 	ctx    context.Context
-	cancel context.CancelFunc
 	method *Method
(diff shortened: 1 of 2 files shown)


lthibault (dismissed):
LGTM. Merging.

lthibault:
Whoops -- just realized the tests are failing.  Is this expected?

zenhack:
Fixed.

lthibault (approved):

PR #260, opened 2022-07-12
Client: hide pointer indirection.

zenhack:
This changes the APIs so that everything operates on Client instead of
*Client. This will smooth over a couple things I've run into, mainly
around being able to implement interfaces where a method needs to
re-assign the receiver.

I'm open to debate as to whether this is a good change. Copying my message from matrix for posterity:

> @lthibault: How do you feel about changing the API to use Client instead of *Client everywhere (and probably hide the pointer indirection inside the struct)? I think this would make a lot of stuff esp. with generics cleaner.
>
> E.g. on my generics branch right now I have:
> 
> ```go
> package capnp
> 
> // The TypeParam interface must be satisified by a type to be used as a canpproto
> // type parameter. This is satisified by all capnproto pointer types. T should
> // be instantiated by the type itself; in type parameter lists you will typically
> // see parameter/constraints like T TypeParam[T].
> type TypeParam[T any] interface {
> 	// Convert the receiver to a Ptr, storing it in seg if it is not
> 	// already associated with some message (only true for Clients and
> 	// wrappers around them.
> 	EncodeAsPtr(seg *Segment) Ptr
>
>	// Decode the pointer as the type of the receiver. Generally,
>	// the reciever will be the zero value for the type.
>	DecodeFromPtr(p Ptr) T
> }
> ```
>
> ...but I'd like to drop the T parameter so it would just be:
>
> ```go
> type TypeParam interface {
>    EncodeAsPtr(*Segment) Ptr
>    DecodeFromPtr(Ptr)
> }
> ```
>
> ...but I can't do the latter because I would have to implement DecodeFromPtr on **Client, since it needs to modify the pointer. But you can't define methods on double-indirect pointers like that.
> Returning the T is kindof an awful hack so we can just define it on *Client instead.
> But if we hide the pointer indirection inside the struct, then we can define the method on *Client, and it can modify it in place.

Despite the size of the diff, I don't think this will be *horribly* disruptive to downstream code; for ocap-md I just needed to re-run the schema compiler and it just worked. Changes should be pretty mechanical in the cases where you are working with raw `Client`s.

### DIFF
--- a/capability_test.go
+++ b/capability_test.go
@@ -117,19 +117,19 @@ func TestReleasedClient(t *testing.T) {
 func TestNullClient(t *testing.T) {
 	ctx := context.Background()
 	c, p := NewPromisedClient(new(dummyHook))
-	p.Fulfill(nil)
+	p.Fulfill(Client{})
 	tests := []struct {
 		name string
-		c    *Client
+		c    Client
 	}{
-		{"nil", nil},
+		{"nil", Client{}},
 		{"promised nil", c},
 	}
 
 	for _, test := range tests {
 		c := test.c
 		t.Run(test.name, func(t *testing.T) {
-			if NewClient(nil) != nil {
+			if (NewClient(nil) != Client{}) {
 				t.Error("NewClient(nil) != nil")
 			}
 			if !c.IsSame(c) {
--- a/answer.go
+++ b/answer.go
@@ -73,7 +73,7 @@ type Promise struct {
 }
 
 type clientAndPromise struct {
-	client  *Client
+	client  Client
 	promise *ClientPromise
 }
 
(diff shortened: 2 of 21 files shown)


zenhack:
Looks like the tests are passing again here too.

capability_test.go
```diff
@@ -117,19 +117,19 @@ func TestReleasedClient(t *testing.T) {
 func TestNullClient(t *testing.T) {
```
lthibault, line 117: I think we can/should run this test in parallel.  Since the actual tests are happening in the context of a `t.Run`, we should also call `t.Helper()` for the top-level test.

lthibault (changes requested):

capability_test.go
zenhack: It seems like that probably applies to most if not all of the tests in this file, and this isn't something I added -- so do we maybe instead want to do a separate PR where we just go through and add t.Parallel() where appropriate?

lthibault: SGTM

lthibault (approved):
LGTM

PR #265, opened 2022-07-13
Assorted clean-up tasks

lthibault:
This PR bundles a few different clean-up chores and minor updates.

- [x] remove unused `needsEscape` and `hexDigit` functions from package `encoding/text`.
- [x] fix (hitherto unencountered) nil-pointer dereference in `TestBuildNodeMap`.
- [x] declare `Helper` tests in `packed` to reduce visual noise in test results and stack traces.
- [x] run message encode/decode tests in parallel (in an attempt to reduce CI timeouts)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -13,5 +13,6 @@ require (
 	github.com/davecgh/go-spew v1.1.0 // indirect
 	github.com/philhofer/fwd v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	golang.org/x/text v0.3.7 // indirect
 	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c // indirect
 )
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -13,5 +13,6 @@ require (
 	github.com/davecgh/go-spew v1.1.0 // indirect
 	github.com/philhofer/fwd v1.1.1 // indirect
 	github.com/pmezard/go-difflib v1.0.0 // indirect
+	golang.org/x/text v0.3.7 // indirect
```
zenhack: It seems silly to me to pull in a new library just to avoid a bug in the stdlib that doesn't actually affect our code; none of these names have unicode punctuation in them, and I don't really forsee that changing, so this seems like it's just going to bloat binaries with no upside? Haven't looked into the implementation, but particiularly since we're not doing a lot of text processing elsewhere in the repo, I wouldn't be surprised if this ended up pulling in a bunch of large unicode tables we aren't otherwise using.

lthibault: Fair point.  I'll revert this now.

zenhack (approved):

Issue #259, opened 2022-07-11
Message should io.WriterTo and io.ReaderFrom


lthibault:
As per #257 this effort would avoid the need to instantiate an encoder/decoder for single messages.

@zenhack My interpretation of [your comment](https://github.com/capnproto/go-capnproto2/issues/257#issuecomment-1175237862) is that this is pure syntactic sugar, and so I would expect these methods to instantiate and call an encoder/decoder under the hood.  Have I correctly understood your intent?

zenhack:
Yes, that's what I had in mind.

PR #266, opened 2022-07-15
Implement io.WriterTo for Message

lthibault:
Partially addresses #259.  

@zenhack `ReadFrom` is a bit tricker to implement because `Decoder.Decode` returns a `*Message`, and there's no obvious place to hook into.  I'll leave that for a separate PR, but any insight would be most welcome.

### DIFF
--- a/message.go
+++ b/message.go
@@ -355,6 +355,23 @@ func alloc(s *Segment, sz Size) (*Segment, address, error) {
 	return s, addr, nil
 }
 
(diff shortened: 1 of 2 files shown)


zenhack (approved):

zenhack:
Re: ReadFrom, you should just be able to assign the result to the receiver's referent, no?

```go
func (m *Message) ReadFrom(...) ... {
    ...
    msg, err := dec.Decode()
    ...
    *m = *msg
    ...
}
```

PR #268, opened 2022-07-22
Bugfix/nil exc check fails

lthibault:
Fixes two error-handling bugs that were generating false positives.  The first is a classic "gotcha" of using `fmt.Errorf`'s `%w` directive; wrapping a nil error in this way will produce a non-nil error.  The second is a classic "gotcha" regarding interfaces; passing a nil `*Exception` as as a parameter of type `error` produces a non-nil `error` interface.

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -234,7 +234,9 @@ func (ans *answer) sendReturn() (releaseList, error) {
 
 	var err error
 	ans.exportRefs, err = ans.c.fillPayloadCapTable(ans.results, ans.resultCapTable)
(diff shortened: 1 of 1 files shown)


zenhack (approved):

zenhack:
Just curious: were you observing the effects of this somewhere, or did you just spot it in the code while working on something else?

lthibault:
@zenhack Sorry for the late response (was traveling).  I observed this while trying to run `BenchmarkPingPong`.  The `ErrorReporter` was incorrectly flagging a bunch of nil errors/exceptions.

PR #273, opened 2022-07-29
Remove server.Policy.

lthibault:
- [x] Remove `server.Policy`
- [x] Remove `*server.Policy` from call sites
- [x] Update code generator

This PR requires you to run `./regen.sh`.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -38,7 +38,7 @@ func (errorEchoImpl) Echo(_ context.Context, call air.Echo_echo) error {
 
 func TestServerCall(t *testing.T) {
 	t.Run("Success", func(t *testing.T) {
-		echo := air.Echo_ServerToClient(echoImpl{}, nil)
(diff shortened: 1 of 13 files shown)


server/server_test.go
```diff
@@ -105,32 +105,13 @@ func (seq *callSeq) GetNumber(ctx context.Context, call air.CallSequence_getNumb
 	return nil
 }
 
-type lockCallSeq struct {
```
zenhack, line 108: Is this just an unrelated drive-by "oh look, this is dead code," or is this related the the primary change in this pr in some way I'm missing?

zenhack (approved):
LGTM.

server/server_test.go
lthibault: Correct, it's just dead code that was flagged by the linter.  I figured it wouldn't add too much noise to the PR (and then forgot to mention it).

PR #274, opened 2022-07-30
First steps towards generics

zenhack:
This adds a `TypeParam` interface capturing operations we will need in order to support generics, and makes sure that the expected types implement it.

Not super useful on its own, but in the interest of keeping PRs small and easy to review...

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -21,6 +21,15 @@ type Interface struct {
 	cap CapabilityID
 }
 
+// i.EncodeAsPtr is equivalent to i.ToPtr(); for implementing TypeParam.
+// The segment argument is ignored.
+func (i Interface) EncodeAsPtr(*Segment) Ptr { return i.ToPtr() }
+
+// DecodeFromPtr(p) is equivalent to p.Interface(); for implementing TypeParam.
+func (Interface) DecodeFromPtr(p Ptr) Interface { return p.Interface() }
+
+var _ TypeParam[Interface] = Interface{}
+
 // NewInterface creates a new interface pointer.
 //
 // No allocation is performed in the given segment: it is used purely
@@ -571,6 +580,18 @@ func (c Client) Release() {
 	<-h.done
 	h.Shutdown()
 }
+
+func (c Client) EncodeAsPtr(seg *Segment) Ptr {
+	capId := seg.Message().AddCap(c)
+	return NewInterface(seg, capId).ToPtr()
+}
+
+func (Client) DecodeFromPtr(p Ptr) Client {
+	return p.Interface().Client()
+}
+
+var _ TypeParam[Client] = Client{}
+
 
 // isResolve reports whether ch has been resolved.
 // The caller must be holding onto ch.mu.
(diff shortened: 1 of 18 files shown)


capnpc-go/templates/interfaceClient
```diff
@@ -30,3 +30,11 @@ func (c {{$.Node.Name}}) AddRef() {{$.Node.Name}} {
 func (c {{$.Node.Name}}) Release() {
 	c.Client.Release()
 }
+
+func (c {{$.Node.Name}}) EncodeAsPtr(seg *{{.G.Capnp}}.Segment) {{.G.Capnp}}.Ptr {
```
zenhack, line 34: Mentioned in the commit message, but: these generated wrappers are necessary for capabilities but not for structs, I think because the generated struct types *embed* `capnp.Struct`, whereas the interface types have an explicitly named field called `Client`. I don't seen an obvious reason why these should be different, so maybe we should just change this so that `Client` is embedded as well? I'm inclined to suggest that as a separate PR though, to keep things orthogonal.

lthibault: SGTM.  Since we look at codegen relatively rarely, I'm going to open your comment as a new issue so that we don't forget.

zenhack:
Just tacked on one more commit, which makes Ptr implement TypeParam as well. But probably should not have a giant generics pr, so once you're happy we should merge.

capability.go
```diff
@@ -572,6 +582,18 @@ func (c Client) Release() {
 	h.Shutdown()
 }
 
+func (c Client) EncodeAsPtr(seg *Segment) Ptr {
+	capId := seg.Message().AddCap(c)
+	return NewInterface(seg, capId).ToPtr()
+}
+
+func (c *Client) DecodeFromPtr(p Ptr) {
```
lthibault, line 590: A side-effect of using pointer receivers is that we will not be able to assign concrete types to `TypeParam`.  It seems like most functions in this library try to avoid returning pointers, so I'm worried this might make composition a bit awkward.

The radical solution would be to deal with `*Ptr`, `*Struct`, `*List` and `*Client` everywhere, but I suspect there's a good reason we avoided that in the first place.

zenhack: Ack, you're right. I suppose we could go back to the solution I had before, where the interface has a type parameter that's supposed to match the receiver, and you have to write `T TypeParam[T]` everywhere, but then decode can take the raw type, rather than a pointer...

I wish I had a way to write generic functions like:

```go
func Foo[*T TypeParam](...) T
```

i.e. specify a constraint that `*T` implements TypeParam, but still have T be the type parameter...

You *can* return a pointer by doing:

```go
type TPPtr[A any] interface {
	TypeParam
	~*A
}

func Foo[A any, PtrA TPPtr[A]](p Ptr) A {
	var a A
	b := PtrA(&a)
	b.DecodeFromPtr(p)
	return a
}

// ...

Foo[A, *A](p)
```

...but that's pretty horrible...

zenhack: Ok, I did a second pass where I tweaked the interface to go back to returning the value, which gets rid of the pointer receiver. A bit more convoluted, but will make call sites better, and it's not really used except by generated code.

zenhack:
Rebased on top of main, dropped a bit of generated code rendered unnecessary by the embedding.

lthibault (approved):

capability.go
lthibault: 😚👌

PR #276, opened 2022-07-30
Embed capnp.Client in generated interface types.

### LINKED ISSUE #275: Embed capnp.Client in capabilities.
>Mentioned in the commit message, but: these generated wrappers are necessary for capabilities but not for structs, I think because the generated struct types *embed* `capnp.Struct`, whereas the interface types have an explicitly named field called `Client`. I don't seen an obvious reason why these should be different, so maybe we should just change this so that `Client` is embedded as well? I'm inclined to suggest that as a separate PR though, to keep things orthogonal.

_Originally posted by @zenhack in https://github.com/capnproto/go-capnproto2/pull/274#discussion_r933763612_

zenhack:
Like we do with struct types. Fixes #275

### DIFF
--- a/capnpc-go/templates/interfaceClient
+++ b/capnpc-go/templates/interfaceClient
@@ -1,7 +1,7 @@
 {{with .Annotations.Doc -}}
 // {{.}}
 {{end -}}
-type {{.Node.Name}} struct { Client {{.G.Capnp}}.Client }
+type {{.Node.Name}} struct { {{.G.Capnp}}.Client }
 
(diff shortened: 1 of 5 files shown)


lthibault (approved):

zenhack:
Test failure looks like the same intermittent issue we're seeing on main?

lthibault:
Agreed.

zenhack:
Ok. I'm going to go ahead and merge then.

PR #278, opened 2022-07-31
Remove a couple uses of the *_List type aliases.

zenhack:
This is in preparation for removing the aliases entirely. Most
importantly, this patch removes the one use inside the code generator
itself.

We can't actually delete the aliases until the code generator no longer
uses them at call sites, but this is the first step.


lthibault (approved):

zenhack:
The failure is one of our known intermittent issues, so I'm going to go ahead and merge.

PR #279, opened 2022-07-31
Kill the struct wrapper around Client

zenhack:
...in generated interface types. This means that Client and the wrapper
types have the same underlying type, so we can CapList for both.

Notes:

- The pogs package has some test failures that need fixing.
- Annoyingly, we can't just use ~Client as the constraint, since the
  proper underlying type is `struct { *client }`. But it works.

---

Marking this as a draft for now.


lthibault (dismissed):

lthibault (approved):

zenhack:
Fixed the tests. One thing I *really* don't like about this is that go doc shows the constraint as `~struct { *client }`, which is exposing implementation detail. And if you do `go doc Client`, it doesn't tell you that the underlying type is the same.

lthibault:
Forgive me if we've covered this before, but is there a reason why the constraint can't be `~Client`?

zenhack:
If I use `~Client`, I get:

```
./list.go:1151:16: invalid use of ~ (underlying type of Client is struct{*client})
```

zenhack:
Per discussion via matrix, I pushed another commit that makes an alias `type ClientKind = ~struct { *client }`, which makes this a bit better.

pogs/pogs_test.go
```diff
@@ -161,13 +161,13 @@ var goodTests = []Z{
...
 	{Which: air.Z_Which_echo, Echo: air.Echo{}},
-	{Which: air.Z_Which_echo, Echo: air.Echo{Client: capnp.ErrorClient(errors.New("boo"))}},
+	{Which: air.Z_Which_echo, Echo: air.Echo(capnp.ErrorClient(errors.New("boo")))},
 	{Which: air.Z_Which_echoes, Echoes: []air.Echo{
-		{Client: capnp.Client{}},
-		{Client: capnp.ErrorClient(errors.New("boo"))},
-		{Client: capnp.Client{}},
-		{Client: capnp.ErrorClient(errors.New("boo"))},
-		{Client: capnp.Client{}},
+		air.Echo{},
```
lthibault, line 166: Nit:  I think this can be simplified to `{}`.  Some linters complain about this, so I vote to reduce such noise.

lthibault (approved):
Minor issue inline, otherwise LGMT.

lthibault (approved):

PR #281, opened 2022-08-01
Clean up formatting, inefassign and misspellings.

lthibault:
Fixes https://github.com/capnproto/go-capnproto2/issues/280

### DIFF
--- a/std/capnp/rpc.capnp
+++ b/std/capnp/rpc.capnp
@@ -645,7 +645,7 @@ struct Disembargo {
   #
   # Message sent to indicate that an embargo on a recently-resolved promise may now be lifted.
   #
-  # Embargoes are used to enforce E-order in the presence of promise resolution.  That is, if an
+  # Embargos are used to enforce E-order in the presence of promise resolution.  That is, if an
   # application makes two calls foo() and bar() on the same capability reference, in that order,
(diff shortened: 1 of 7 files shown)


std/capnp/rpc.capnp
```diff
@@ -645,7 +645,7 @@ struct Disembargo {
   #
   # Message sent to indicate that an embargo on a recently-resolved promise may now be lifted.
   #
-  # Embargos are used to enforce E-order in the presence of promise resolution.  That is, if an
+  # Embargoes are used to enforce E-order in the presence of promise resolution.  That is, if an
```
zenhack, line 648: I think I would prefer *not* to fix spelling errors in this file manually, since it's copied over from the reference implementation, and is going to get blown away whenever we pull in changes from upstream. Instead, if these still exist upstream, you could submit a patch. If not, they'll get fixed when we pull in the versions from the next release.

zenhack (changes requested):

std/capnp/rpc.capnp
lthibault: Good point, I hadn't realized these were upstream files.  I'll just revert this change.

zenhack (approved):

zenhack:
Failure looks like one of the intermittent ones we've seen before; going to assume not related and merge.

PR #285, opened 2022-08-05
Add client methods to interfaceClient template

lthibault:
As discussed on Matrix, commonly-used `Client` methods are not accessible from subtypes of `Client`.

This PR adds the following methods to the `interfaceClient` template:

- [x] `client.String`
- [x] `client.Resolve`
- [x] `client.SetFlowLimiter`

It also adds docstrings to existing methods.

@zenhack What are your thoughts the other methods (`WeakRef`, `State`, `IsSame`, ...)?  I can't quite decide whether it makes sense to export these.  Are they deliberately part of the public API, or are they intended for internal use, but originally exported out of necessity?

### DIFF
--- a/capnpc-go/templates/interfaceClient
+++ b/capnpc-go/templates/interfaceClient
@@ -26,7 +26,7 @@ func (c {{$.Node.Name}}) {{.Name|title}}(ctx {{$.G.Imports.Context}}.Context, pa
 // should not be used to compare clients.  Use IsSame to compare clients
 // for equality.
 func (c {{$.Node.Name}}) String() string {
-	return fmt.Sprintf("%T(%v)", c, capnp.Client(c))
+	return {{$.G.Imports.Fmt}}.Sprintf("%T(%v)", c, capnp.Client(c))
 }
 
 // AddRef creates a new Client that refers to the same capability as c.
--- a/capnpc-go/templates/interfaceClient
+++ b/capnpc-go/templates/interfaceClient
@@ -45,6 +45,12 @@ func (c {{$.Node.Name}}) Release() {
 	capnp.Client(c).Release()
 }
 
+// Resolve blocks until the capability is fully resolved or the Context
+// expires.
+func (c {{$.Node.Name}}) Resolve(ctx context.Context) error {
+	return capnp.Client(c).Resolve(ctx)
+}
+
 func (c {{$.Node.Name}}) EncodeAsPtr(seg *capnp.Segment) capnp.Ptr {
 	return capnp.Client(c).EncodeAsPtr(seg)
 }
--- a/capnpc-go/templates/interfaceClient
+++ b/capnpc-go/templates/interfaceClient
@@ -66,8 +66,8 @@ func (c {{$.Node.Name}}) IsValid() bool {
 	return capnp.Client(c).IsValid()
 }
 
-// IsSame reports whether c and c2 refer to a capability created by the
-// same call to NewClient.  This can return false negatives if c or c2
+// IsSame reports whether c and other refer to a capability created by the
+// same call to NewClient.  This can return false negatives if c or other
 // are not fully resolved: use Resolve if this is an issue.  If either
 // c or other are released, then IsSame panics.
 func (c {{$.Node.Name}}) IsSame(other {{$.Node.Name}}) bool {
--- a/capnpc-go/templates/interfaceClient
+++ b/capnpc-go/templates/interfaceClient
@@ -21,10 +21,18 @@ func (c {{$.Node.Name}}) {{.Name|title}}(ctx {{$.G.Imports.Context}}.Context, pa
 }
 {{end}}
 
+// AddRef creates a new Client that refers to the same capability as c.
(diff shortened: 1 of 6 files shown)


capnpc-go/templates/interfaceClient
```diff
@@ -21,14 +21,36 @@ func (c {{$.Node.Name}}) {{.Name|title}}(ctx {{$.G.Imports.Context}}.Context, pa
 }
 {{end}}
 
+// String returns a string that identifies this capability for debugging
+// purposes.  Its format should not be depended on: in particular, it
+// should not be used to compare clients.  Use IsSame to compare clients
+// for equality.
+func (c {{$.Node.Name}}) String() string {
+	return capnp.Client(c).String()
```
zenhack, line 29: I think it would be a bit better to return something that reflects the type of the capability, rather than just shelling out to untyped Client. Maybe`fmt.Sprintf("%T(%v)", c, capnp.Client(c))`

capnpc-go/templates/interfaceClient
```diff
@@ -37,6 +59,17 @@ func ({{$.Node.Name}}) DecodeFromPtr(p capnp.Ptr) {{$.Node.Name}} {
...
+// been released.
 func (c {{$.Node.Name}}) IsValid() bool {
 	return capnp.Client(c).IsValid()
 }
+
+// Update the flowcontrol.FlowLimiter used to manage flow control for
+// this client. This affects all future calls, but not calls already
+// waiting to send. Passing nil sets the value to flowcontrol.NopLimiter,
+// which is also the default.
+func (c {{$.Node.Name}}) SetFlowLimiter(lim flowcontrol.FlowLimiter) {
```
zenhack, line 73: Maybe also expose GetFlowLimiter?

zenhack (changes requested):
Re: other methods:

- IsSame() definitely makes sense.
- I feel ambivalent about State(); it's such a rarely used thing that having to cast to Client might actually be good to keep it out of the namespace for "normal" uses.
- I think WeakRef makes sense, but also I think we should make WeakRef a generic type `type WeakRef[T ~ClientKind]`) if we're going to do that (and also get rid of the pointer indirection like we did with Client. Probably enough stuff be a separate PR.

capnpc-go/templates/interfaceClient
lthibault: I was wondering about that.  Is there any reason the user should want to get a flow limiter?  I was thinking this is more of an internal operation that doesn't necessarily need to be exported.

lthibault:
Agreed for `WeakRef`.  Leaving that for a separate PR.

capnpc-go/templates/interfaceClient
```diff
@@ -66,6 +66,14 @@ func (c {{$.Node.Name}}) IsValid() bool {
 	return capnp.Client(c).IsValid()
 }
 
+// IsSame reports whether c and c2 refer to a capability created by the
```
zenhack, line 69: You named the parameter `other` below, so this probably should use that name instead of `c2`.

capnpc-go/templates/interfaceClient
```diff
@@ -26,7 +26,7 @@ func (c {{$.Node.Name}}) {{.Name|title}}(ctx {{$.G.Imports.Context}}.Context, pa
 // should not be used to compare clients.  Use IsSame to compare clients
 // for equality.
 func (c {{$.Node.Name}}) String() string {
-	return capnp.Client(c).String()
+	return {{$.G.Imports.Fmt}}.Sprintf("%T(%v)", c, capnp.Client(c))
```
zenhack, line 29: I wonder if we should actually just remove these indirections as well, like we did for `{{.G.Capnp}}`. But that doesn't need to be this pr.

zenhack (changes requested):
Almost there.

capnpc-go/templates/interfaceClient
lthibault: d'oh!

capnpc-go/templates/interfaceClient
lthibault: I suspect the reason for this is to avoid unneeded imports.  For instance, if your schema doesn't contain any capabilities, there's no need to import "fmt".

capnpc-go/templates/interfaceClient
zenhack: I guess it just seems odd to expose one and not the other. But I don't have *that* strong of an opinion, so I'll defer if you feel strongly.

lthibault: Nah, I'm just being lazy.  I'll add it.

zenhack (approved):

capnpc-go/templates/interfaceClient
zenhack: Ah. I think it would be simpler to (unconditionally) just emit a block like:

```
var (
  _ fmt.Print
  ...
)
```

...for each package that might or might not be used. I think this is how the protobuf compiler does things.

PR #286, opened 2022-08-05
Clean up import rendering in codegen.

lthibault:
This is a small refactor that aims for an incremental improvement to `fileparts.go`'s readability.

I had to stare at the `imports` struct a wee bit longer than I would have liked, owing to the indirection in its initialization.  This PR removes the `imports.init` method and replaces it with a simple `newImports` constructor.  Runtime behavior is unaffected.

⚠️ N.B.:  this is branched off of #285, so merge that one first.

### DIFF
--- a/capnpc-go/fileparts.go
+++ b/capnpc-go/fileparts.go
@@ -76,12 +76,14 @@ type imports struct {
 	used  map[string]bool // keyed on import path
 }
 
-func newImports() (i imports) {
+func newImports() imports {
+	i := imports{used: make(map[string]bool)}
+
 	for _, spec := range importList {
 		i.reserve(spec)
 	}
 
-	return
+	return i
 }
 
 func (i *imports) Capnp() string {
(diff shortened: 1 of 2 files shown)


zenhack:
Hm, this still has the commits from #285 on it, even though that's merged... makes the diff hard to follow...

lthibault:
Rats.  I'll rebase and push.

lthibault:
Fixed.

lthibault:
Bump.  Is this okay to merge?

capnpc-go/fileparts.go
```diff
@@ -140,6 +149,10 @@ func (i *imports) byName(name string) (spec importSpec, ok bool) {
 }
 
 func (i *imports) add(spec importSpec) (name string) {
+	if i.used == nil {
+		i.used = make(map[string]bool)
```
zenhack, line 153: Why do this here, as opposed to just putting it at the top of newImports?

lthibault: No particular reason.  I think this is an artifact from the refactoring where there was some lazy initialization going on.

I'll make the change.

lthibault: Fixed.

zenhack (approved):

Issue #272, opened 2022-07-29
Bootstrap with expired capability deadlocks


lthibault:
Passing an expired capability to `rpc.Conn.Bootstrap` causes a deadlock.  To reproduce, run `TestHandleReturn_regression` in the `bug/bootstrap-panic` branch.  The second subtest relates to the present issue.

```go
func TestDeadlock(t *testing.T) {
	p1, p2 := transport.NewPipe(1)

	srv := testcp.PingPong_ServerToClient(pingPongServer{}, nil)
	conn1 := rpc.NewConn(rpc.NewTransport(p2), &rpc.Options{
		BootstrapClient: srv.Client,
	})
	defer conn1.Close()

	conn2 := rpc.NewConn(rpc.NewTransport(p1), nil)
	defer conn2.Close()

	// create an EXPIRED context
	ctx, cancel := context.WithCancel(context.Background())
	cancel()

	// NOTE: bootstrap with expired context
	pp := testcp.PingPong{Client: conn2.Bootstrap(context.Background())}
	defer pp.Release()

	f, release := pp.EchoNum(ctx, func(ps testcp.PingPong_echoNum_Params) error {
		ps.SetN(42)
		return nil
	})
	defer release()

	_, err := f.Struct()  // DEADLOCK
	assert.NoError(t, err)
}
```
```
panic: test timed out after 10s

goroutine 50 [running]:
testing.(*M).startAlarm.func1()
	/usr/local/go/src/testing/testing.go:2029 +0x8e
created by time.goFunc
	/usr/local/go/src/time/sleep.go:176 +0x32

goroutine 1 [chan receive]:
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1405 +0x4a5
testing.tRunner(0xc0001076c0, 0xc000165c60)
	/usr/local/go/src/testing/testing.go:1445 +0x13b
testing.runTests(0xc0001350e0?, {0x14549e0, 0x18, 0x18}, {0x100e3e5?, 0x1014806?, 0x145a000?})
	/usr/local/go/src/testing/testing.go:1837 +0x457
testing.(*M).Run(0xc0001350e0)
	/usr/local/go/src/testing/testing.go:1719 +0x5d9
capnproto.org/go/capnp/v3/rpc_test.TestMain(0x1040531?)
	/Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/level0_test.go:43 +0xbb
main.main()
[...]

Issue #271, opened 2022-07-29
Method call with expired context causes nil-pointer panic


lthibault:
To reproduce:  run `TestHandleReturn_regression` in `level0_test.go`, under the `bug/bootstrap-panic` branch.

```go
func TestPanic(t *testing.T) {
	p1, p2 := transport.NewPipe(1)

	srv := testcp.PingPong_ServerToClient(pingPongServer{}, nil)
	conn1 := rpc.NewConn(rpc.NewTransport(p2), &rpc.Options{
		BootstrapClient: srv.Client,
	})
	defer conn1.Close()

	conn2 := rpc.NewConn(rpc.NewTransport(p1), nil)
	defer conn2.Close()

	pp := testcp.PingPong{Client: conn2.Bootstrap(context.Background())}
	defer pp.Release()

	// create an EXPIRED context
	ctx, cancel := context.WithCancel(context.Background())
	cancel()

	f, release := pp.EchoNum(ctx, func(ps testcp.PingPong_echoNum_Params) error {
		ps.SetN(42)
		return nil
	})
	defer release()

	_, err := f.Struct()  // BANG!
	assert.NoError(t, err)
}
```

```
panic: Promise.Fulfill called after Fulfill or Reject

goroutine 15 [running]:
capnproto.org/go/capnp/v3.(*Promise).Fulfill(0x0?, {0xc000132570, 0x0, 0x0, {0x0, 0x0}, 0x0, 0x80})
	/Users/lthibault/Go/src/capnproto.org/go/capnp/answer.go:132 +0x135
capnproto.org/go/capnp/v3/rpc.(*Conn).handleReturn.func2.2()
	/Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/rpc.go:952 +0x53
capnproto.org/go/capnp/v3/internal/syncutil.Without({0x12f3fb0, 0xc000120cd8}, 0xc000049f60)
	/Users/lthibault/Go/src/capnproto.org/go/capnp/internal/syncutil/with.go:19 +0x8c
capnproto.org/go/capnp/v3/rpc.(*Conn).handleReturn.func2()
	/Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/rpc.go:951 +0x21e
created by capnproto.org/go/capnp/v3/rpc.(*Conn).handleReturn
	/Users/lthibault/Go/src/capnproto.org/go/capnp/rpc/rpc.go:947 +0x697
[...]

PR #288, opened 2022-08-08
Fix 271

zenhack:
The actual fix is the last commit; see the commit message.

There are also a couple other improvements in this pr that I made along the way:

1. Just above the fix for #271, there was a spot where we were mutating the questions table without holding c.mu. I have not observed symptoms from this, but this is wrong so I added a syncutil.With around it while I was there.
2. The panic message for this failure mode is now more informative; it tells you which method was called originally, and if it was reject, what the error was.

Note, this fixed *both* tests; I can't reproduce #272. @lthibault, are you still able to reproduce it on this branch?


zenhack:
While debugging this, I added some code to record a stack trace for the first call to Fulfill/Reject, which was helpful. I didn't keep it in because it means keeping a buffer around for each fulfilled promise, and the way I did it I just oversized it so it would be 1MiB per promise, but worth bearing in mind/maybe thinking about how we can make the system easier to introspect for stuff like this.

lthibault (approved):
That was quick! 😃 

I'll run the tests on my machine after merging.

PR #290, opened 2022-08-16
.gitignore: ignore test coverage profiles

zenhack:
...this is a common name for them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@
 TAGS
 *~
 *.swp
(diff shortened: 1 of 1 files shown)


lthibault (dismissed):
Perhaps consider `*.out`?  That will catch things like `mem.out` and `cpu.out` for pprof, and other such artifacts.

lthibault (approved):

zenhack:
Done.

PR #294, opened 2022-08-21
Add BBR Flow-Control

zenhack:
This is still WIP; what's left:

- [ ] Finish up the various state logic (state.go)
- [ ] Test everything thoroughly and fix all the bugs
- [x] Expose stuff via the flowcontrol package
- [ ] Misc. cleanup
- [x] ~~Probably clean up the history~~ we're just going to squash-merge this.

Marking it as a draft, but the basic structure is there.


internal/bbr/filter.go
```diff
@@ -0,0 +1,128 @@
...
+
+func newRtPropFilter() rtPropFilter {
+	return rtPropFilter{
+		nextSample: rtPropSample{
+			// Set this to a value that will be removed immediately,
+			// and will be lower priority than any other value, so
+			// it will removed from the queue when a new sample is
+			// added.
+			//
+			// XXX: this still could theoretically do the wrong thing
```
lthibault, line 52: Would such a failure be transient, or would rtProp be permanently stuck in a bad state?

internal/bbr/filter.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+func (f *rtPropFilter) AddSample(sample rtPropSample) {
+	// We want to avoid an un-bounded growing queue for two reasons:
+	//
+	// 1. Space usage
+	// 2. Recomputing the estimate is an O(n) operation, so we want to
+	//    keep n small if we're doing that on each ack.
+	//
+	// We manage this as follows: rather than adding each sample to the queue
```
lthibault, line 74: I'm having a bit of trouble parsing this sentence.  Rather than add each sample to the queue, we do what?

internal/bbr/manager.go
```diff
@@ -0,0 +1,255 @@
...
+		panic("TODO: overflow")
+	}
+	replyChan := make(chan packetMeta)
+	select {
+	case <-ctx.Done():
+		return func() {}, ctx.Err()
+	case <-m.ctx.Done():
+		return func() {}, m.ctx.Err()
+	case m.chSend <- sendRequest{size: int64(size), replyChan: replyChan}:
+		pm := <-replyChan
```
lthibault, line 41: Are we sure this can never block?  Do we need to select against `ctx.Done()`?

internal/bbr/readme.go
```diff
@@ -0,0 +1,5 @@
+// package bbr implements the BBR congestion control algorithm, as
+// described in:
+//
+// https://queue.acm.org/detail.cfm?id=3022184
+package bbr
```
lthibault, line 5: Maybe rename the file to the standard `doc.go`?

internal/clock/clock.go
```diff
@@ -0,0 +1,131 @@
...
+
+import (
+	"math"
+	"sync"
+	"time"
+
+	"capnproto.org/go/capnp/v3/internal/syncutil"
+)
+
+// A Clock can measure time.
```
lthibault, line 16: You don't say! 😆 😜

lthibault (commented):
Just skimmed this and left a couple of comments/questions inline.   Hope this is helpful.

internal/bbr/filter.go
zenhack: This is stale; originally I had the value as zero but for other reasons I changed it to MaxInt64, so this will be superseded on the first sample in all cases now. Fixed the docs.

internal/bbr/manager.go
zenhack: Yes we're sure. I added some comments to the chSend field to clarify how this works.

internal/bbr/filter.go
zenhack: I made a stab at clarifying it; does the new version make any more sense?

internal/bbr/readme.go
zenhack: Done.

internal/clock/clock.go
zenhack: I think this was a case of "having trouble thinking of what to put -> prototyping mode, fix it later."

lthibault: Of course - just teasing 😉

internal/bbr/filter.go
lthibault: Much clearer.  I have just one remaining grammatical nit that makes me backtrack a few times while reading.  What do you think of this?

>We manage this as follows: rather than adding each sample to the queue individually, we compute the minimum RTT for each 1-second window, and add that aggregate value to the queue, which in turn drops samples more than 30 seconds old. This gives the benefits of the sliding window without needing to store each and every sample.

zenhack: LGTM, done.

lthibault (approved):

zenhack:
Status update:

- Most of the algorithm is implemented. What's left is some of the details of the probeRTT state, which are a bit ambiguous to me from the paper -- I plan to look at the actual kernel implementation to get clarification on those points.
- The tests indicate that the estimate of the bottleneck bandwidth is sometimes too low; I need to debug this.
- The implementation needs more rigorous testing, and of course any bugs that shakes out will need to be fixed.

flowcontrol/bbr/state.go
```diff
@@ -0,0 +1,180 @@
...
+	// Our bandwidth estimate as of the last time we received an ack:
+	prevBtlBwEstimate bytesPerNs
+
+	// How many times in a row we've tried increasing sending rate and
+	// not seen a signficant bandwidth increase:
+	plateauRounds int
+}
+
+func (s *startupState) initialize(lim *Limiter) {
+	lim.cwndGain = 2 / math.Ln2
```
lthibault, line 28:
```
// cwndGain (Congestion WiNDow; standard terminology) is a factor
// by which we may exceed BDP. i.e. if inflight() is greater than
// estimated BDP * cwndGain, attempts to send will block.
```


Sanity check:  we do not exceed BDB unless inflight > ~2.89 ?

(Under what conditions is it necessary to exceed BDP again?)

zenhack: In order to test whether our current estimate is too low, we have to try sending faster than that estimate would recommend, to see if we get better bandwidth that way (if so, the estimate was too low, if not, it wasn't). So cwndGain gives us some headroom to do those experiments.

Not sure what your first question is asking? `inflight()` returns the number of bytes currently on the wire (an integer).

lthibault: Makes sense 👍
First question was just verifying that the threshold seemed reasonable.  I forget why, but it caused me some confusion this morning.

zenhack: Yeah, it's about 2.89; the idea is we want it to increase exponentially with each round trip. The paper claims this is the minimum value that will facilitate that, but tbh I don't understand why 2 doesn't work.

Should I go ahead and merge this then, and we'll deal with the remaining issues as a follow up?

lthibault: 👍
And yes, go for it!

PR #311, opened 2022-09-27
Fix #306

### LINKED ISSUE #306: Use after free on importClients
When attempting to upload a large directory with ocap-merkledag, I see an imported capability being released, and then subsequently passed to another call, resulting an exception due to the invalid export id; this is indicative of a bug in go-capnp, and @lthibault suggested that this might be the source of a problem he'd observed in wetware as well. There is a formatted trace of the entire run here:

https://mirror.zenhack.net/pub/trace.log

Unfortunately it is not a particularly minimal example.

Notes on the format of the trace:

- The `"Send"` or `"Recv"` tag indicates whether the message being logged was sent or received; this is recorded from the client's perspective.
- The content fields in the call/return payloads have been nulled out, because they are relatively uninteresting and include some large byte blobs which aren't interesting for our purposes.

zenhack:
Also, document the fact that ClientPromise.Fulfill steals the reference.

Mixing this up was the root cause of the issue. Specifically both
Promise.ReleaseClients() and Message.Reset() were releasing the same
clients, causing them to be invalid if later used.

This patch just adds a call to .AddRef() in the appropriate spot, so
that the promise has its own references to the clients, as the API
suggests.

---

N.B. this patch is stacked on top of #310 to avoid conflicts; that PR should be reviewed & merged first.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -199,7 +199,7 @@ func (p *Promise) resolve(r Ptr, e error) {
 			res := resolution{p.method, r, e}
 			for path, cp := range p.clients {
 				t := path.transform()
-				cp.promise.Fulfill(res.client(t))
+				cp.promise.Fulfill(res.client(t).AddRef())
 				cp.promise = nil
 			}
 			if p.callsStopped != nil {
--- a/answer.go
+++ b/answer.go
@@ -60,8 +60,8 @@ type Promise struct {
 	// clients is a table of promised clients created to proxy the eventual
 	// result's clients.  Even after resolution, this table may still have
 	// entries until the clients are released. Cannot be read or written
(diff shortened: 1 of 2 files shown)


zenhack:
Test suite is hanging; I got so excited when it fixed ocap-md I forgot to run the tests :sweat_smile:

Marking as a draft while I figure out what's wrong.

zenhack:
Looks like the test failures are due to release messages not being sent, which is not a surprising symptom of a regression -- so I still have some debugging to do. Will look at it more tomorrow.

zenhack:
I'm going to close this one. The ownership questions could use better documentation, but this is not the correct solution.

PR #317, opened 2022-10-01
Fix 315

zenhack:
Notes:

- This is stacked on top of #316, which should be reviewed & merged first.
- I have not properly tested this, beyond making sure it doesn't break existing tests, so I'm going to mark it as a draft and find time to do the tests in the next day or two. But it's not complicated and "should" work, if you're desperate to try it. I'm at least confident it's not *more* broken.

The last commit message discusses the solution in more detail.

### DIFF
--- a/rpc/question.go
+++ b/rpc/question.go
@@ -125,6 +125,13 @@ func (q *question) handleCancel(ctx context.Context) {
 		}
 		close(q.finishMsgSend)
 
+		// First remove the question from the table, so we don't
+		// double-resolve the promise. We can't free the id though,
+		// since the finish failed.
+		syncutil.With(&q.c.mu, func() {
+			q.c.questions[q.id] = nil
+		})
+
 		q.p.Reject(rejectErr)
 		if q.bootstrapPromise != nil {
 			q.bootstrapPromise.Fulfill(q.p.Answer().Client())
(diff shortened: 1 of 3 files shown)


lthibault (approved):

lthibault (approved):

zenhack:
Alright, I added a test, which passes. Marking this ready for review.

zenhack:
Hm, not sure what the failure is about; tests pass repeatedly on my machine.

lthibault (approved):

lthibault:
**tl;dr** Maybe just check that `err != nil` instead of asserting a specific error?

@zenhack The [failing test](https://github.com/capnproto/go-capnproto2/actions/runs/3178376659/jobs/5179794227#step:5:1467) looks to be a race condition between the connection being closed and the expired context being detected by the bootstrap call.  As we're explicitly testing for the `context.Canceled` error, we get a failure when conn is closed first.

It seems to me we only care that _some_ error was returned when passing an expired context, not the specific error.

(P.S.:  passes on my machine, too.)

zenhack:
@lthibault, hm, should it actually be possible to get a connection drop in that case? Why would that be happening in that test?

lthibault:
> @lthibault, hm, should it actually be possible to get a connection drop in that case? Why would that be happening in that test?

I'd have to look again, but I think the connection is closed if bootstrap fails.

zenhack:
Per discussion on IRC, I'm going to merge and open a separate issue about the flaky tests

Issue #318, opened 2022-10-09
Flaky RPC tests


zenhack:
There are a few tests in the RPC package that usually pass, but fail intermittently:

- [x] `TestRecvCancel` (hangs)
- [ ] `TestSendCancel` (maybe; see https://github.com/capnproto/go-capnproto2/issues/318#issuecomment-1276920615)
- [x] `TestHandleReturn_regression` (Sometimes hangs, sometimes actually fails, see https://github.com/capnproto/go-capnproto2/issues/318#issuecomment-1276866593 )
- [x] `TestBootstrapReceiverAnswerRpc`: see https://github.com/capnproto/go-capnproto2/issues/318#issuecomment-1276839048
- [x] `TestRecvAbort`:

```
level0_test.go:148: 
        	Error Trace:	level0_test.go:148
        	Error:      	Received unexpected error:
        	            	context deadline exceeded
        	Test:       	TestRecvAbort
        	Messages:   	should resolve bootstrap capability
```

There may be one or two more. These crop up running the tests locally often enough to be easily reproducible but not every time, so at some point they managed to slip past CI and get merged. We should fix these; I don't want to keep having to wonder when working on something else whether I've broken something or a test is just flaky; this slows down development across the board.

zenhack:
Observations:

- `TestRecvCancel` only ever hangs when run in parallel with other tests. It passes consistently if you either do `go test -run TestRecvCancel` or delete the `t.Parallel()` at the top of the function. I don't yet understand why that would happen; it doesn't look like there are any shared resources with other tests that might interfere, and there's no timeouts or anything...
- The timeout in TestRecvAbort trigger the above failure isn't theoretically impossible, though a 1 second timeout seems generous. I'm having trouble reproducing the issue right now, but if I was able to reproduce it semi-reliably, my first thought would be to try commenting out t.Parallel() to see if that fixes anything.

It seems likely that some of these are race conditions that are sufficiently timing dependent that the noise from other tests is needed to trigger them.

---

Also, I managed to trigger a failure other than the above:

```
--- FAIL: TestBootstrapReceiverAnswerRpc (0.00s)
panic: test.capnp:CapArgsTest.call: rpc: connection closed [recovered]
	panic: test.capnp:CapArgsTest.call: rpc: connection closed

goroutine 27 [running]:
testing.tRunner.func1.2({0x667820, 0xc000586060})
	/usr/lib/go/src/testing/testing.go:1396 +0x24e
[...]

PR #320, opened 2022-10-13
Better isolate the tests under TestHandleReturn_regression

zenhack:
Previously, these tests used the same connection for both tests. This fixes intermittent test failures on the second test, as discussed in #318.

I think, in principle, the tests *should* pass in that case anyway, but:

1. Keeping the cases isolated makes this easier to understand
2. This way we can (and now do) run the tests in parallel
3. While we should better test the scenarios where:
   - a connection is used after a bootstrap fails
   - bootstrap is called more than once
   - etc. Doing so without tests dedicated to those things will be very difficult to maintain; we could add some but that should be a separate (low priority) task.

### DIFF
--- a/rpc/level0_test.go
+++ b/rpc/level0_test.go
@@ -1772,52 +1772,62 @@ func TestSendCancel(t *testing.T) {
 func TestHandleReturn_regression(t *testing.T) {
 	t.Parallel()
 
-	p1, p2 := transport.NewPipe(1)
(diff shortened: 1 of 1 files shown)


lthibault (approved):
LGTM.

zenhack:
The failure is for a test I didn't touch, and I think I've seen that one flake before, so I should probably add it to the list of things to investigate for #318. Merging this one then.

PR #322, opened 2022-10-14
Fix a regression introduced by releaseQuestions()

zenhack:
This was manifesting as occasional failures of
TestBootstrapReceiverAnswerRpc, as discussed in #318. When this was introduced, I attempted to avoid double-rejecting the promise by removing it from the table after the first rejection, but this is in fact incorrect in the case where we cancel the question, because the entry needs to stick around until the return message comes in.

With this patch, instead, we solve the problem by just having releaseQuestions() check the `finished` flag before calling Reject.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -399,7 +399,13 @@ func (c *Conn) releaseAnswers(answers map[answerID]*answer) {
 
 func (c *Conn) releaseQuestions(questions []*question) {
 	for _, q := range questions {
-		q.Reject(ExcClosed)
+		canceled := q != nil && q.flags&finished != 0
+		if !canceled {
+			// Only reject the question if it isn't already flagged
+			// as finished; otherwise it was rejected when the finished
+			// flag was set.
+			q.Reject(ExcClosed)
+		}
 	}
 }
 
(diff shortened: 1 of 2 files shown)


rpc/rpc.go
```diff
@@ -399,7 +399,13 @@ func (c *Conn) releaseAnswers(answers map[answerID]*answer) {
 
 func (c *Conn) releaseQuestions(questions []*question) {
 	for _, q := range questions {
-		q.Reject(ExcClosed)
+		canceled := q != nil && q.flags&finished != 0
```
lthibault, line 402: Minor thing that totally shouldn't be fixed now:  maybe add methods to a `flags` type, so that we can do `q.flags.Finished()`?

lthibault (approved):
LGTM.  Based on your second comment, sounds like you're inclined to merge it.  Totally fine by me, but I'll let you push the button just in case.

rpc/rpc.go
zenhack: That seems reasonable. Or maybe at least a flags.Contains(finished), which would be less boilerplate to implement.

PR #323, opened 2022-10-14
Add Go 1.19 to the build matrix.

zenhack:


zenhack:
Both of the failures I see in CI are among the tests listed for #318, so they should not block merging.

lthibault (approved):

Issue #301, opened 2022-09-03
Data race detected in TestBootstrapReceiverAnswerRpc


zenhack:
See: https://github.com/capnproto/go-capnproto2/runs/8171454309?check_suite_focus=true#step:5:1504

This is intermittently causing CI to fail; I'm fairly sure the changes to the docs are not the source of the race in that PR :P

PR #324, opened 2022-10-14
Server.handleCall: Run methods when the ctx is already canceled

zenhack:
...and let the method implementation check the context itself.

This solves an occasional hang (observed in #318) in TestRecvCancel, which involves a method implementation that waits for its context to be canceled and then closes a channel. Without this patch, if the context is cancelled early enough for handleCall to see it, it won't run the method at all, and so the channel will never be closed, causing the test to hang on a receive on that channel.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -203,10 +203,7 @@ func (srv *Server) handleCalls(ctx context.Context) {
 func (srv *Server) handleCall(ctx context.Context, c *Call) {
 	defer srv.wg.Done()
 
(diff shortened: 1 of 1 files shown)


lthibault (approved):
👍 I like this a lot.  In addition to simplifying the code, there have been times where I've wanted to detect that a call was made and aborted by the caller.

zenhack:
The data race seems to be the same as the one mentioned in #301, so I'm going to go ahead and merge.

PR #326, opened 2022-10-14
Conn.sendMessage: don't return an error.

### LINKED ISSUE #304: Handle errors returned from `Conn.sendMessage`
While investigating #303, @zenhack and I observed that most calls to `Conn.sendMessage` were ignoring the returned error.  The immediately problematic call sites have been fixed, but we'll want to respond to those other errors as well.  At a minimum, we might consider passing them to `ErrorReporter`, or terminating the connection.

@zenhack suggested it might be possible to detect this issue using a custom linter.

zenhack:
Instead, arrange for the error handling logic in callback() to always be run.

Fixes #304, by obviating the need for extra checks.

---

I came to this when I realized a missing check could be the source of the intermittent `TestRecvAbort` failures. Sadly, it was not the cause, but this is still worth doing.

### DIFF
--- a/rpc/import.go
+++ b/rpc/import.go
@@ -98,9 +98,8 @@ func (ic *importClient) Send(ctx context.Context, s capnp.Send) (*capnp.Answer, 
 	q := ic.c.newQuestion(s.Method)
 
 	// Send call message.
-	var err error
 	syncutil.Without(&ic.c.mu, func() {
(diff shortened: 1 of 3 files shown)


lthibault (approved):

zenhack:
That's one of the known flaky tests; merging.

PR #327, opened 2022-10-17
Fix intermittent hang in TestRecvAbort

zenhack:
This fixes a problem where, if a bootstrap message is in the send queue but not sent when the connection begins shutdown, both connection shutdown and resolution of the bootstrap's promise will deadlock, which manifests as a timeout in TestRecvAbort.

The issue is that Bootstrap() starts a task to keep the connection from shutting down while it is sending the message, and stops it in the latter callback to sendMessage() -- but that callback may not be run until drainQueue is executed, which happens *after* stopTasks() returns, hence the deadlock.

This whole part of the design is a bit sketchy; we also invoke question.handleCancel in said callback, which itself invokes sendMessage(). We should audit for correctness of the design; I expect we'll find some bugs.

For the present bug though, there's no actual reason why the task needs to be held open until the message is actually sent; we can close it once the message is in the queue, which is what question.PipelineSend() does, so now Conn.Bootstrap() does the same. At some point we should factor out some of the common code between them.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -207,6 +207,7 @@ func (c *Conn) Bootstrap(ctx context.Context) (bc capnp.Client) {
 	if !c.startTask() {
 		return capnp.ErrorClient(rpcerr.Disconnectedf("connection closed"))
 	}
+	defer c.tasks.Done()
 
 	bootCtx, cancel := context.WithCancel(ctx)
 	q := c.newQuestion(capnp.Method{})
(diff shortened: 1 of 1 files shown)


lthibault:
>This whole part of the design is a bit sketchy; we also invoke question.handleCancel in said callback, which itself invokes sendMessage(). We should audit for correctness of the design; I expect we'll find some bugs.

Can we create a card for this?  Sounds like a low-hanging opportunity to improve robustness.

zenhack:
I opened #329; if that's what you want me to work on next I can.

PR #333, opened 2022-11-14
Return capnp.Client from promised :Capability type

lthibault:
This PR updates codegen for `:Capability` types so that futures return `capnp.Client` instead of `capnp.Future`.

Along the way, common code for handling `:AnyPointer` has been abstracted.  See `any_pointer.go`.

Once merged, users will need to run `./regen.sh`.


capnpc-go/capnpc-go.go
```diff
@@ -410,6 +409,14 @@ func (g *generator) Value(rel *node, t schema.Type, v schema.Value) (string, err
 	}
 }
 
+func isAnyCap(ap schema.Type_anyPointer) bool {
+	unconstrained := ap.
+		Which() == schema.Type_anyPointer_Which_unconstrained
+	capability := ap.Unconstrained().
```
zenhack: Is it ok to do this without first verifying `unconstrained == true`? My memory is this should panic if it's another variant, but I could be misremembering.

zenhack (commented):
This seems broadly reasonable, modulo one thing mentioned in-line.

capnpc-go/capnpc-go.go
lthibault: It should be safe; we were doing it in an equivalent way before.  I believe union types will panic, but this is a group.

zenhack:
The data race looks like #301 to me; I'm ok merging if you are.

PR #335, opened 2022-11-14
Add code generation for AnyList

lthibault:
This PR adds code generation for the `AnyList` type.  To do so, it adds a `capnp.Future.List()` method that is analogous to `capnp.Future.Struct()`.

Note:  in a future effort, we should probably rename the `Struct()` method on generated `*_Future` types to something like `Wait()`.

### DIFF
--- a/capnpc-go/templates/promiseFieldAnyList
+++ b/capnpc-go/templates/promiseFieldAnyList
@@ -1,0 +1,3 @@
+func (p {{ .Node.Name }}_Future) {{ .Field.Name|title }}() *capnp.Future {
+	return  p.Future.Field({{ .Field.Slot.Offset }}, nil)
+}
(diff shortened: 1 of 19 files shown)


capnpc-go/templates/promiseFieldAnyList
```diff
@@ -0,0 +1,3 @@
+func (p {{ .Node.Name }}_Future) {{ .Field.Name|title }}() *capnp.Future {
```
zenhack, line 1: I feel like we should define and return a List_Future type, rather than continuing to just use plain Future; that's sortof the point of specializing AnyList in the first place, no?

lthibault: As per discussion on Matrix, I am merging now and we can add this in a separate PR.

PR #336, opened 2022-11-14
Add missing accessor for promised AnyStruct

lthibault:
⚠️ Merge after #335


capnpc-go/templates/promiseFieldAnyStruct
```diff
@@ -0,0 +1,3 @@
+func (p {{ .Node.Name }}_Future) {{ .Field.Name|title }}() *capnp.Future {
```
zenhack, line 1: Same thing I pointed out in #335; I feel like we should add a wrapper type Struct_Future or such.

lthibault: As per discussion on Matrix, I am merging now and we can add this in a separate PR.

PR #338, opened 2022-11-17
Rename server.Call.Ack() -> .Go()

### LINKED ISSUE #289: Rename server.Ack() -> server.Go()
This came up when @lthibault and I were discussing flow control; `.Ack()` is confusing in that it's not sending an acknowledgement or anything like it; rather, it spawns a goroutine to handle further calls. Before we stabilize the API I'd like to change it to a better name. Louis suggested `.Go()`, which I like.

zenhack:
Also improve some of the docs involving it.

Fixes #289

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -53,17 +53,18 @@ func (c *Call) AllocResults(sz capnp.ObjectSize) (capnp.Struct, error) {
 	return c.results, err
 }
 
-// Ack is a function that is called to acknowledge the delivery of the
-// RPC call, allowing other RPC methods to be called on the server.
-// After the first call, subsequent calls to Ack do nothing.
+// Go is a function that is called to unblock future calls; by default
+// a server only accepts one method call at a time, waiting until
+// the method returns before servicing the next method in the queue.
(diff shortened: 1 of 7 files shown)


server/server.go
```diff
@@ -53,17 +53,18 @@ func (c *Call) AllocResults(sz capnp.ObjectSize) (capnp.Struct, error) {
 	return c.results, err
 }
 
-// Ack is a function that is called to acknowledge the delivery of the
-// RPC call, allowing other RPC methods to be called on the server.
-// After the first call, subsequent calls to Ack do nothing.
+// Go is a function that is called to unblock future calls; by default
+// a server only accepts one method call at a time, waiting until
+// the method returns before servicing the next method in the queue.
+// calling Go spawns another goroutine to service additional calls
```
lthibault, line 59: Capital `C`.

lthibault (changes requested):

server/server.go
zenhack: Done.

lthibault (approved):

PR #346, opened 2022-11-24
Add benchmark for streaming calls.

zenhack:
Note, this includes the fix from #345, which should be merged first.

As expected, the profile output has runtime.mallocgc at the top.

### DIFF
--- a/rpc/bench_test.go
+++ b/rpc/bench_test.go
@@ -6,9 +6,55 @@ import (
 	"testing"
 
 	"capnproto.org/go/capnp/v3"
+	"capnproto.org/go/capnp/v3/flowcontrol"
 	"capnproto.org/go/capnp/v3/rpc"
 	testcp "capnproto.org/go/capnp/v3/rpc/internal/testcapnp"
+	"capnproto.org/go/capnp/v3/std/capnp/stream"
 )
+
+func BenchmarkStreaming(b *testing.B) {
+	ctx := context.Background()
+	p1, p2 := net.Pipe()
+	srv := testcp.StreamTest_ServerToClient(nullStream{})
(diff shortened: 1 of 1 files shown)


zenhack:
Ack, just realized this doesn't really measure what I wanted it to; marking as a draft for the moment.

zenhack:
It's hard to see how that failure could be related to adding a new benchmark, so I'm going to assume this was a pre-existing bug that is rare enough for it to have slipped by so far; I'll open an issue.

rpc/bench_test.go
```diff
@@ -2,18 +2,64 @@ package rpc_test
...
+	bootstrap := testcp.StreamTest(conn2.Bootstrap(ctx))
+	defer bootstrap.Release()
+	var (
+		futures      []stream.StreamResult_Future
+		releaseFuncs []capnp.ReleaseFunc
+	)
+	bootstrap.SetFlowLimiter(flowcontrol.NewFixedLimiter(1 << 9))
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		for j := 0; j < 1<<16; j++ {
```
lthibault, line 34: Just in case you haven't already checked, `go test -bench` usually does a pretty good job at picking a large enough `b.N`.  Perhaps we don't need the nested `for` loop?

lthibault (approved):

rpc/bench_test.go
zenhack: The thing is, b.N looks for the smallest value that gives consistent measurement. here that's not actually what we want -- we need it to send enough messages for re-use to matter, whereas if we just go off b.N, we may not really be using the pool.

lthibault: Makes sense -- just double-checking 👍

zenhack:
Ok, that's a *different* panic, opened #349 for it. Going to run again.

PR #347, opened 2022-11-24
transport: Use a sync.Pool in NewMessage() 

zenhack:
(stacked on top of #346; merge that first).

This cuts down on allocations a lot, see:

```
$ go tool pprof before.mem
rpc.test
Type: alloc_space
Time: Nov 23, 2022 at 9:43pm (EST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 632.75MB, 76.25% of 829.88MB total
Dropped 47 nodes (cum <= 4.15MB)
Showing top 10 nodes out of 70
      flat  flat%   sum%        cum   cum%
  209.20MB 25.21% 25.21%   209.20MB 25.21%  capnproto.org/go/capnp/v3.(*MultiSegmentArena).Allocate
   86.01MB 10.36% 35.57%   115.01MB 13.86%  capnproto.org/go/capnp/v3/rpc/transport.(*ctxReader).Read
   85.51MB 10.30% 45.88%   122.51MB 14.76%  capnproto.org/go/capnp/v3/rpc/transport.(*ctxWriteCloser).write
   66.01MB  7.95% 53.83%    66.01MB  7.95%  net.(*pipeDeadline).set
   50.51MB  6.09% 59.92%    57.01MB  6.87%  capnproto.org/go/capnp/v3.NewPromise (inline)
   47.51MB  5.72% 65.64%   173.52MB 20.91%  capnproto.org/go/capnp/v3.(*Decoder).Decode
      29MB  3.49% 69.14%   238.20MB 28.70%  capnproto.org/go/capnp/v3.NewMessage
      23MB  2.77% 71.91%       23MB  2.77%  capnproto.org/go/capnp/v3/internal/chanmutex.NewLocked (inline)
   20.50MB  2.47% 74.38%   258.71MB 31.17%  capnproto.org/go/capnp/v3/rpc/transport.(*transport).NewMessage
   15.50MB  1.87% 76.25%   175.08MB 21.10%  capnproto.org/go/capnp/v3/rpc.(*Conn).handleCall

$ go tool pprof after.mem
rpc.test
Type: alloc_space
Time: Nov 23, 2022 at 9:43pm (EST)
Entering interactive mode (type "help" for commands, "o" for options)
(pprof) top
Showing nodes accounting for 457.05MB, 72.35% of 631.73MB total
Dropped 45 nodes (cum <= 3.16MB)
Showing top 10 nodes out of 68
      flat  flat%   sum%        cum   cum%
[...]

Issue #349, opened 2022-11-27
panic (slice out of bounds) in tests


zenhack:
https://github.com/capnproto/go-capnproto2/actions/runs/3537277747/jobs/5979526799#step:7:742

```
 panic: runtime error: slice bounds out of range [:72] with capacity 0

goroutine 58 [running]:
capnproto.org/go/capnp/v3.(*Segment).slice(...)
	/home/runner/work/go-capnproto2/go-capnproto2/segment.go:49
capnproto.org/go/capnp/v3.(*Segment).readUint64(...)
	/home/runner/work/go-capnproto2/go-capnproto2/segment.go:65
capnproto.org/go/capnp/v3.(*Segment).readRawPointer(...)
	/home/runner/work/go-capnproto2/go-capnproto2/segment.go:69
capnproto.org/go/capnp/v3.(*Segment).resolveFarPointer(0x908dd80, 0x40)
	/home/runner/work/go-capnproto2/go-capnproto2/segment.go:235 +0xa38
capnproto.org/go/capnp/v3.(*Segment).readPtr(0x908dd80, 0x40, 0xffffffff)
	/home/runner/work/go-capnproto2/go-capnproto2/segment.go:116 +0x4b
capnproto.org/go/capnp/v3.Struct.Ptr({0x908dd80, 0x40, {0x0, 0x1}, 0xffffffff, 0x0}, 0x0)
	/home/runner/work/go-capnproto2/go-capnproto2/struct.go:109 +0x10c
capnproto.org/go/capnp/v3.Transform({0x908dd80, 0x40, 0x0, {0x0, 0x1}, 0xffffffff, 0x0}, {0x90111e0, 0x1, 0x1})
	/home/runner/work/go-capnproto2/go-capnproto2/answer.go:579 +0x17b
capnproto.org/go/capnp/v3.resolution.ptr({{0xa7317bd7216570aa, 0x9, {0x0, 0x0}, {0x0, 0x0}}, {0x908dd80, 0x40, 0x0, {0x0, ...}, ...}, ...}, ...)
	/home/runner/work/go-capnproto2/go-capnproto2/answer.go:605 +0x7b
capnproto.org/go/capnp/v3.resolution.client({{0xa7317bd7216570aa, 0x9, {0x0, 0x0}, {0x0, 0x0}}, {0x908dd80, 0x40, 0x0, {0x0, ...}, ...}, ...}, ...)
[...]

PR #350, opened 2022-11-27
Use a sync.Pool for MultiSegmentArenas

zenhack:
Superseeds #347. note that relative to #347 this does a bit more work in cleanup to make sure memory is properly zeroed and we don't mis-report the number of segments when we get a new arena, though these weren't caught in testing on the other PR.

Benchmark results are similar.

While working on this, I caught a bug in the transport tests: the test assume that it is safe to call the ReleaseFunc returned from NewMessage multiple times, which was incidentally true before but undocmented. This patch documents this invariant, and also ensures that it holds even when release() actually does something with the arena.

### DIFF
--- a/message.go
+++ b/message.go
@@ -110,7 +110,7 @@ func NewSingleSegmentMessage(b []byte) (msg *Message, first *Segment) {
 	return msg, first
 }
 
-// Analogous to NewSingleSegmentMessage, but using MutliSegment.
+// Analogous to NewSingleSegmentMessage, but using MultiSegment.
 func NewMultiSegmentMessage(b [][]byte) (msg *Message, first *Segment) {
 	msg, first, err := NewMessage(MultiSegment(b))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -480,9 +480,43 @@ type MultiSegmentArena [][]byte
...
+		return multiSegmentPool.Get().(*MultiSegmentArena)
+	}
+	return multiSegment(b)
+}
+
+// Return this arena to an internal sync.Pool of arenas that can be
+// re-used. Any time MultiSegment(nil) is called, arenas from this
+// pool will be used if available, which can help reduce memory
+// allocations. Calling Release however is optional; if not done
+// the garbage collector will release the memory per usual.
```
lthibault, line 493: Consider documenting that segments are zeroed, for the security-conscious user.

zenhack:
Failure looks to be #349 again

message.go
```diff
@@ -480,9 +480,43 @@ type MultiSegmentArena [][]byte
...
+		for j, _ := range v {
+			v[j] = 0
+		}
+
+		// Truncate the segment, since we use the length as the marker for
+		// what's allocated:
+		(*msa)[i] = v[:0]
+	}
+	(*msa) = (*msa)[:0] // Hide the segments
+	multiSegmentPool.Put(msa)
```
lthibault, line 506: A frequent issue with `sync.Pool` is that large objects may be kept alive despite not being used to their fullest capacity, resulting in high memory usage.  The following mitigation strategies are common:

1. drop objects larger than some arbitrary size,
2. maintain separate pools for objects of similar size, and;
3. try to estimate the "drop" threshold dynamically ([example](https://pkg.go.dev/github.com/libp2p/go-buffer-pool))

We may wish to kick this particular can down the road, but we should probably open an issue to track it.  Perhaps in the interim, we might use strategy 1 with a generous cutoff?

rpc/transport/transport.go
```diff
@@ -147,7 +153,17 @@ func (s *transport) NewMessage(ctx context.Context) (_ rpccp.Message, send func(
 		return err
 	}
 
-	return rmsg, send, func() { msg.Reset(nil) }, nil
+	release = func() {
+		if alreadyReleased {
```
lthibault, line 157: This won't protect us from concurrent calls to `release()`. Is that a problem?

Go 1.19's `atomic.Bool.Swap(true)` might be a good fit, here.

lthibault (approved):
Approving, as this looks to be correct.  See inline for a few questions and suggestions.

message.go
lthibault: Clarification:  strategy 2 has the effect of increasing the probability of large objects being GCed.  I've seen it mixed and matched with strategies 1 & 3.

zenhack: In the interim, I don't think it's likely to be a problem for RPC workloads, since there's going to be a cap on message sizes anyway. But yeah, we could open an issue to track this stuff.

lthibault: Addendum:  we might also consider applying these strategies on a per-segment basis.

lthibault: Ah yeah, that's a good point.  Great!  👍 

Resolving.

message.go
zenhack: Done.

lthibault (approved):

message.go
zenhack: Opened #351 for this.

rpc/transport/transport.go
zenhack: It's not a problem for the places it's actually used (which afaik is just the tests). It feels wrong to introduce atomics for something production code doesn't actually rely on...

lthibault: SGTM.  Resolving.

zenhack:
Failure looks like #349 again

PR #296, opened 2022-08-23
simplify transport.

zenhack:
A substantial amount of the complexity of the transport package is due
the interface: we pass contexts to functions that do IO, and then jump
through hoops to respect those contexts. Instead, this patch changes
Transport and Codec's Close() to work like net.Close(): it interrupts
any outstanding IO. We demand the same semantics from the
io.ReadWriteCloser that we pass in, which seems like not a tall order.

Still TODO:

- [ ] rework the rpc subsystem's shutdown logic. The basic idea is we
spawn an extra goroutine to wait on context.Done() and then call close,
but the shutdown logic is intricate and I am too tired right now; this
patch introduces test regressions for now.
- [ ] Tweak the docs on how this changes the interface for e.g. NewConn

---

Marking this as a draft for now, until I find time to tie up the loose ends.

PR #354, opened 2022-11-28
Rpc: split IO into dedicated goroutines.

zenhack:
This is a self-contained enough change that it seemed worth submitting independently of #296; this moves the actual IO calls on the transport into a new pair of goroutines (one for send, one for receive), so that the receive loop and the send goroutine can always be responsive to context cancels. This will become important when we no longer require the IO operations themselves to respond to a context.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -12,6 +12,7 @@ import (
 	"capnproto.org/go/capnp/v3/exc"
 	"capnproto.org/go/capnp/v3/exp/mpsc"
 	"capnproto.org/go/capnp/v3/internal/syncutil"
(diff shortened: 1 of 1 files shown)


zenhack:
For once I think the data race in the tests might be a new bug; marking as a draft until I'm sure.

zenhack:
Ok, I pushed a variation on this that only splits off receives into a new goroutine, which I think is all we actually need; the data race was because the new writer goroutine was still running during shutdown, so instead we just rely on the existing send logic to return in a reasonable amount of time; unlike reads we don't expect it to block indefinitely.

lthibault (approved):
LGTM.

PR #355, opened 2022-11-30
Add helpers Serve and ListenAndServe

hspaay:
First draft of a helper for Serve and ListenAndServe.
Using context for ListenAndServe as it is the only way to close it when needed.


lthibault:
Just a quick thing:  can you lowercase the file names?

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"net"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+
+	"capnproto.org/go/capnp/v3/rpc"
+)
+
+//clientHook := {}
```
lthibault: dead code?

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+
+func TestServe(t *testing.T) {
```
lthibault: Can we mark this test as parallel, using `t.Parallel()`?

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+	lis, err := net.Listen("tcp", ":8888")
```
lthibault: Let's use `:0` so that this test doesn't fail if something is already bound to `8888`?

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+	assert.NoError(t, err)
+
+	errChannel := make(chan error)
+	go func() {
+		err2 := rpc.Serve(lis, &rpc.Options{})
```
lthibault: You can pass `nil` instead of `&rpc.Options{}`.

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+		errChannel <- err2
+	}()
+	time.Sleep(time.Second)
```
lthibault: This is a good place for a select statement.  Recommend rewriting as:

```go
select {
case <-time.After(time.Second * 2):
   // fail the test
case err := <-errChannel:
    assert.ErrorIs(t, err, net.ErrClosed)
}
```

rpc/Serve_test.go
lthibault: We should also `defer lis.Close()` in case something goes wrong in the test, so that we can release the port.

rpc/Serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+func TestListenAndServe(t *testing.T) {
+	var err error
+	ctx, cancelFunc := context.WithCancel(context.Background())
+	errChannel := make(chan error)
+	go func() {
+		err2 := rpc.ListenAndServe(ctx, ":8888", &rpc.Options{})
+		errChannel <- err2
+		t.Log("Serve has ended")
+	}()
+	time.Sleep(time.Second)
```
lthibault: See above.

lthibault (changes requested):
This is definitely on the right track.  Don't be alarmed by the number of suggestions -- I'm just being thorough 🙂 

And by the way, thanks for the contribution! 🙌

rpc/serve_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}()
+	time.Sleep(time.Second)
+	t.Log("Closing listener")
+	err = lis.Close()
+	assert.NoError(t, err)
+	err = <-errChannel
+	assert.ErrorIs(t, err, net.ErrClosed)
+}
+
+func TestListenAndServe(t *testing.T) {
```
lthibault, line 33: Oh, and let's mark this one with `t.Parallel()` as well. 

In fact, most of the comments from `TestServe` apply here as well.

rpc/Serve_test.go
hspaay: yep, leftover of some experimentation

rpc/Serve_test.go
hspaay: Oh I had to look that one up. :0 is the next available port. Nifty!

hspaay:
What is still missing is TLS support. I think that is a must for use in production.

rpc/serve.go
```diff
@@ -0,0 +1,49 @@
...
+	for {
+		rwc, err := lis.Accept()
+		if err != nil {
+			return err
+		}
+		// For each new incoming connection, create a new RPC transport connection that will serve incoming RPC requests
+		// rpc.Options will contain the bootstrap capability
+		go func() {
+			transport := NewStreamTransport(rwc)
+			conn := NewConn(transport, opt)
```
zenhack, line 21: One subtle thing here: you are going to want to make a copy of the options, cloning the bootstrap interface with `.AddRef()`. The reason for this is that the connection takes ownership of the bootstrap interface, so will drop it when the connection exits. So you want each connection to have its own reference. And you'll also want to drop the original reference when Serve() returns.

rpc/serve.go
```diff
@@ -0,0 +1,49 @@
...
+
+			<-conn.Done()
+			// Remote client connection closed
+			return
```
zenhack, line 25: I don't think you need this return statement.

rpc/serve.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+// ListenAndServe opens a listener on the given address and serves a Cap'n Proto RPC to incoming connections
+// network and address are passed to net.Listen. Use network "unix" for Unix Domain Sockets
+// and "tcp" for regular TCP connections.
+func ListenAndServe(ctx context.Context, network, addr string, opt *Options) error {
+	//var listener net.Listener
+	//var err error
+
+	//listener, err = net.Listen(network, address)
```
zenhack, line 37: Should delete the commented-out bits.

zenhack (changes requested):
A few things in-line. Additionally, it would be good to have a test or two that supplies a bootstrap interface and checks that it behaves as expected, incl. across connections (would catch the issue I pointed out in Serve()).

zenhack:
Quoting Henk (2022-11-29 21:24:50)

>    On that last note, I found that when passing context to ...
>    "ListenConfig).Listen(ctx, " ... it doesn't seem to respond to
>    cancelling the context. Not sure if there is a good reason for using
>    context here, other than as convention.

If there's no actual effect then I'd prefer just using net.Listen; keep
it simple.

hspaay:
> If there's no actual effect then I'd prefer just using net.Listen; keep it simple.

After digging a bit deeper it is passed to a netFD.connect which does use it although how exactly I can't follow. I'm a bit concerned what happens if the context isn't closed after exit. ... since it isn't doing anything for our use-case lets stick to the regular listen.

rpc/Serve.go
```diff
@@ -0,0 +1,65 @@
...
+// Context can be used to stop listening.
+func ListenAndServe(ctx context.Context, address string, options *Options) error {
+	var listener net.Listener
+	var err error
+
+	if address == "" {
+		return fmt.Errorf("missing address")
+	}
+	// UDS paths start with either '.' or '/'
+	if strings.HasPrefix(address, "unix:") {
```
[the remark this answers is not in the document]
hspaay: After digging a bit deeper, the context is passed to a netFD.connect which does use it although how exactly I can't follow. I'm a bit concerned what happens if the context isn't closed after exit. ... since it isn't doing anything for our use-case lets stick to the regular listen.

rpc/serve.go
hspaay: @zenhack ,
Right, that is a good catch.  That said, Serve doesn't return until the connection is closed. Does this have to take into account that the bootstrap interface is used elsewhere or can be cancelled?

I wonder whether passing a bootstrap interface to the Serve function implies you hand over ownership. It is similar to obtaining a capability and handing it over elsewhere. The application is responsible to invoke AddRef if needed. Isn't that the same here?

zenhack: We should document it but, yeah that was my thinking: Serve should take ownership of the reference, for consistency with what NewConn does.

hspaay:
> A few things in-line. Additionally, it would be good to have a test or two that supplies a bootstrap interface and checks that it behaves as expected, incl. across connections (would catch the issue I pointed out in Serve()).

Indeed a good point. I'll expand the tests with using a bootstrap interface.

lthibault (changes requested):

rpc/serve.go
hspaay: > So you want each connection to have its own reference. And you'll also want to drop the original reference when Serve() returns.
Ohh I missed the point here. This looks like a major issue. So for each new incoming connection, use the Addref on Options.BootstrapClient and release when the  connection closes. Also release when Serve ends itself. Add note to documentation. Got it and will submit an update.

rpc/Serve.go
```diff
@@ -0,0 +1,65 @@
...
+	"context"
+	"fmt"
+	"net"
+	"strings"
+)
+
+// Serve serves a Cap'n Proto RPC to incoming connections
+// Serve exits with the listener error
+func Serve(lis net.Listener, options *Options) error {
+	if options == nil { //|| !options.BootstrapClient.IsValid() {
```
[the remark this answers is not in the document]
hspaay: Hmm, after further consideration, I think a check on the  bootstrap interface being valid is needed. The NewConn might support Nil options but Serve will not function as intended. Rather than let the user figure out why it doesn't work, an error could inform them a valid bootstrap interface is required.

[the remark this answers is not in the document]
lthibault: >The NewConn might support Nil options but Serve will not function as intended.

How do you mean?  Any errors encountered will be reported as errors when you call methods on the bootstrap capability. 

Also, I think it's not as simple as calling `c.IsValid()` since the capability must first resolve.  Unless I'm missing something, I think it's much better to stay "hands off".

rpc/Serve.go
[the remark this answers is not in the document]
lthibault: It's used when binding to the local port or file-descriptor.  Beyond that, it's no longer used, so there's no danger of a resource leak if it isn't canceled.

rpc/Serve.go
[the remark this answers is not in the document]
hspaay: As discussed, changed API to pass the bootstrap client instead of options

rpc/serve.go
hspaay: Updated Serve with AddRef and update the testcase to better check for proper release

rpc/rpc.go
```diff
@@ -124,9 +125,15 @@ type ErrorReporter interface {
...
-// Passing nil for opts is the same as passing the zero value.
+// NewConn creates a new connection that communicates on a given transport.
+//
+// Closing the connection will close the transport and release the bootstrap
+// client provided in opts.
+//
+// Passing nil for opts is the same as passing the zero value. This can be useful
+// for obtaining a reference to the remote side's bootstrap. This in turn allows
+// this listening connection to be used as the RPC client and make calls to the remote
+// side, since capnp connections are symmetric.
```
zenhack, line 136: I think these last two sentences have too little context to be useful, and the potential to be confusing (notably, passing in a bootstrap interface of your own doesn't stop you from calling .Bootstrap() on the connection to get the other side's bootstrap -- notably, Sandstorm does both when launching a grain).

I think there's often a temptation to try to stretch this into explaining more than really fits in the reference docs for one function, but what's really needed is better overview/tutorial docs.

rpc/serve.go
```diff
@@ -0,0 +1,72 @@
...
+func Serve(lis net.Listener, bootstrapClient capnp.Client) error {
+	if !bootstrapClient.IsValid() {
+		err := errors.New("BootstrapClient is not valid")
+		return err
+	}
+	// Accept incoming connections
+	for {
+		rwc, err := lis.Accept()
+		if err != nil {
+			// Since we took ownership of the bootstrap client, release it after we're done.
```
zenhack, line 25: I would instead just put `defer bootstrapClient.Release()` at the top of the function.

rpc/serve.go
```diff
@@ -0,0 +1,72 @@
...
+			bootstrapClient.Release()
+			return err
+		}
+
+		// For each new incoming connection, create a new RPC transport connection that will serve incoming RPC requests
+		// rpc.Options will contain the bootstrap capability
+		go func() {
+			// skip if the bootstrap client has closed since receiving the connection
+			// this can happen if the server exits while incoming connections are made
+			if bootstrapClient.IsValid() {
```
zenhack, line 38: You don't need to check this repeatedly; once at the top is plenty.

rpc/serve.go
```diff
@@ -0,0 +1,72 @@
...
+				transport := NewStreamTransport(rwc)
+				// the RPC connection takes ownership of the bootstrap interface and will release it when the connection
+				// exits, so use AddRef to avoid releasing the provided bootstrap client capability.
+				opts := Options{
+					BootstrapClient: bootstrapClient.AddRef(),
```
zenhack, line 43: We should hoist the call to .AddRef() up above, before we fork off a goroutine. This avoids a scenario where:

- We fork off a goroutine to handle one connection
- The next call to Accept() fails for some reason, so we return, dropping the reference to bootstrapClient
- The goroutine we forked tries to .AddRef(), and it panics since bootstrapClient has already been released.

zenhack (changes requested):
A few more things in line.

rpc/rpc.go
hspaay: Yeah I can see your point. I found this such a valuable bit of insight that I put it in there, but yeah it fits better in documentation.
I've removed it when changing the API to take bootstrapClient instead of Options.

rpc/serve.go
hspaay: The reason I added this was to prevent a panic due to a race condition.
When the server connection is closed and bootstrapClient is released,  the go-routine that handles the last incoming connection can still be running and invoke bootstrapClient.AddRef(). AddRef panics when this happens.
It is a bit of an edge case but might avoid a proper shutdown.

rpc/serve.go
hspaay: Well spotted. This is exactly the problem I ran into and described above. Excellent point and great solution!

rpc/serve.go
```diff
@@ -0,0 +1,69 @@
...
+		err := errors.New("BootstrapClient is not valid")
+		return err
+	}
+	// Accept incoming connections
+	defer bootstrapClient.Release()
+	for {
+		conn, err := lis.Accept()
+		if err != nil {
+			// Since we took ownership of the bootstrap client, release it after we're done.
+			if !bootstrapClient.IsValid() {
```
zenhack, line 27: I don't think you need this check either.

rpc/serve.go
```diff
@@ -0,0 +1,69 @@
...
+		opts := Options{
+			BootstrapClient: bootstrapClient.AddRef(),
+		}
+
+		// For each new incoming connection, create a new RPC transport connection that will serve incoming RPC requests
+		// rpc.Options will contain the bootstrap capability
+		go func() {
+			transport := NewStreamTransport(conn)
+			conn := NewConn(transport, &opts)
+			<-conn.Done()
```
zenhack, line 44: Come to think of it, you don't even need to spawn a goroutine here; instead, just skip the call to .Done() and you can do it in the main loop; it doesn't block.

zenhack (changes requested):
Almost there!

rpc/serve.go
hspaay: Looks like I forgot to release the bootstrap client here. Replacing the check with release.

rpc/rpc.go
```diff
@@ -130,10 +130,7 @@ type ErrorReporter interface {
 // Closing the connection will close the transport and release the bootstrap
 // client provided in opts.
 //
-// Passing nil for opts is the same as passing the zero value. This can be useful
-// for obtaining a reference to the remote side's bootstrap. This in turn allows
-// this listening connection to be used as the RPC client and make calls to the remote
-// side, since capnp connections are symmetric.
+// Passing nil for opts is the same as passing the zero value.
```
lthibault, line 133: Nit:  I know what you mean, but I worry this might be a bit cryptic to a newbie.  How about something like:

> If opts == nil, sensible defaults are used.  See Options.

?

rpc/serve.go
```diff
@@ -19,34 +19,24 @@ func Serve(lis net.Listener, bootstrapClient capnp.Client) error {
...
-				err = errors.New("the bootstrap client was already released")
-			}
 			bootstrapClient.Release()
 			return err
 		}
 
+		// the RPC connection takes ownership of the bootstrap interface and will release it when the connection
+		// exits, so use AddRef to avoid releasing the provided bootstrap client capability.
+		opts := Options{
+			BootstrapClient: bootstrapClient.AddRef(),
```
lthibault, line 34: Nit:  `bootstrapClient` is of type `capnp.Client`, so the `Client` suffix is redundant.  I suggest renaming to `boot`.

rpc/serve.go
```diff
@@ -19,34 +19,24 @@ func Serve(lis net.Listener, bootstrapClient capnp.Client) error {
...
-					BootstrapClient: bootstrapClient.AddRef(),
-				}
-				conn := NewConn(transport, &opts)
-				<-conn.Done()
-				// Remote client connection closed
-			}
-		}()
+		transport := NewStreamTransport(conn)
+		rpcConn := NewConn(transport, &opts)
+		_ = rpcConn
```
lthibault, line 39: What's going on here?

lthibault (changes requested):
Almost done.  The first PR always involves a bit of back-and-forth 🙂

rpc/rpc.go
hspaay: Sounds good.

rpc/serve.go
hspaay: The goroutine is not needed as zenhack pointed out, so the NewConn is all there is. 
The line _ = rpcConn can be removed. It was there to inspect the result with the debugger. I'll fix that.
Btw there is a comment that diff left out:
>	// For each new incoming connection, create a new RPC transport connection that will serve incoming RPC requests
You okay with this?

lthibault: 👍

rpc/serve.go
hspaay: I guess that is a style choice. Personally I prefer more descriptive names to point the casual reader in the right direction. If the style here is to keep things short and not repeat yourself then I'll remove it. Not sure what you guys prefer. (is there a style guide?)

lthibault: The general Go convention is to let the types do (some of) the talking.  The variable name `boot` tells us how the `capnp.Client` is to be used, so it's adding new information.  `bootstrapClient capnp.Client` kind of "stutters", and has the unfortunate effect of lengthening Go's already-quite-long lines.

All this to say that I have a moderate preference for the DRY approach.  But I don't want to die on this hill, either.

hspaay: I'll give it a try, thanks for explaining

rpc/serve.go
zenhack: Hm? that's what the defer is for?

hspaay: Argh!

lthibault (approved):
I think this is good to go! 🙂 🎉

lthibault:
Test failure looks like #349.  Re-running tests.

PR #359, opened 2022-11-30
Fix a bug pertaining to casting struct lists to pointer lists

zenhack:
Yesterday Kenton looped me in on https://capnproto.org/news/2022-11-30-CVE-2022-46149-security-advisory.html so I could see if we were affected too. We aren't, but I found a different (non-security) bug in the relevant area of the code.

See the comment on the test.

The change to the if condition is just formatting; I found it difficult to read when trying to understand it.


lthibault (approved):
Nice catch!

PR #360, opened 2022-12-01
Conn.sender: mpsc -> spsc

zenhack:
Minor followup to #358: We don't actually need multiple-producers here, since we're always holding Conn.mu when we send.


zenhack:
I can't reproduce the failure locally, and it's in a test that doesn't pull in the rpc package as a dependency, so it seems impossible for it to be a result of this patch. I'm going to re-run the tests.

PR #367, opened 2022-12-06
Revert "Fix #301."

### LINKED ISSUE #301: Data race detected in TestBootstrapReceiverAnswerRpc

lthibault:
This reverts commit a2ae8c40e8d8988874ad03bc2980c4c6449f2f7d.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -206,16 +206,12 @@ func (srv *Server) handleCall(ctx context.Context, c *Call) {
 	err := c.method.Impl(ctx, c)
 
 	c.recv.ReleaseArgs()
-	// NOTE: it is important that we call return *before* fulfilling
-	// the answerQueue, otherwise an incoming pipelined call on the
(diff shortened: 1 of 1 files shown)


zenhack (approved):
For posterity: this is the source of #349; the analysis in the (now removed) comment is correct insofar as it identifies what was causing a data race before, but doing it the other way panics, because when the finish message comes in the message will be freed, possibly before answerQueue.fulfill() is done with it, causing a panic.

My next task is to come up with a proper fix for both of these issues, but the consequences of the data race are less bad than the panic, and this is blocking @lthibault from releasing, so we're reverting this for the moment.

PR #368, opened 2022-12-07
Factor out duplicated clientPath type.

zenhack:
This was previously defined privately in both server and the main capnp package.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -3,10 +3,10 @@ package capnp
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"sync"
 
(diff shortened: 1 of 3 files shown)


zenhack:
So actually, the weird thing about this is that the server package doesn't even actually use clientPath as a map key; maybe instead of factoring this out, we can just delete the one in server and use []TransformOp directly instead?

lthibault:
I'm all for simplicity, so +1 to using `[]TransformOp` directly.

zenhack:
Ok, I'm going to close then and I'll open an new one.

PR #369, opened 2022-12-07
Minor readability improvement.

zenhack:
Using a statement as general as switch when a if/else will do makes things marginally harder to understand at a glance.

### DIFF
--- a/server/answer.go
+++ b/server/answer.go
@@ -148,8 +148,7 @@ type queueCaller struct {
 
 func (qc queueCaller) PipelineRecv(ctx context.Context, transform []capnp.PipelineOp, r capnp.Recv) capnp.PipelineCaller {
 	qc.aq.mu.Lock()
-	switch {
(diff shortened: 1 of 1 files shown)


server/answer.go
```diff
@@ -160,7 +159,7 @@ func (qc queueCaller) PipelineRecv(ctx context.Context, transform []capnp.Pipeli
 			return nil
 		}
 		return b.recv(ctx, transform, r)
-	default:
+	} else {
```
lthibault, line 162: It seems like the `else` isn't needed here, since the `if` block returns.  I would personally find it even more readable without the `else`.

lthibault (changes requested):

server/answer.go
zenhack: I agree. Force-pushed a version that makes this change.

lthibault (approved):

PR #370, opened 2022-12-07
Cleanup: add .Contains() method to rpc.answerFlags.

zenhack:
This makes the call sites easier to follow.

We could do the same for questions, but I only want to shave so many yaks for one night.


lthibault (approved):
This has been on my mind as well.  Great minds think alike.  👍

PR #372, opened 2022-12-07
Remove duplicate implementation of clientPath.

zenhack:
This is redundant in two senses:

- It's copypasta from the main package.
- It isn't actually needed, since unlike the main package we never actually use it as a map key. So we can just use []capnp.PipelineOp directly.


lthibault:
This one has a small merge conflict.  Looks like the `default:` case that was removed in #369 .

zenhack:
Fixed the conflict.

zenhack:
Race looks like the one we know about, re-running.

lthibault (approved):

PR #373, opened 2022-12-07
Remove some vestigial bits, fix #352

zenhack:
In particular:

- resultsCapTable has been vestigial since we stopped clearing the
  CapTable on send, and I've been meaning to delete it. Instead, we just
  use the message's cap table directly, which also simplifies some code.
- We no longer need the code that releases Clients from the result's
  CapTable -- this is done by msg.Reset when we free the message. In
  principle, this shouldn't have hurt since Client.Release() is
  idempotent, but the way we were doing it caused the data race in https://github.com/capnproto/go-capnproto2/issues/352
  Instead, we only need to release the additional items in exports if
  releaseResultCaps = true.

---

Note, this is stacked on top of #371, which should be merged first.

PR #374, opened 2022-12-07
Use refcounting for results

zenhack:
Stacked on top of #373.

---

There is no functional change here yet; this just factors the recounting
we were already doing into a package, and exposes the ability to
refcount to other parts of the code.

This is a stepping stone towards addressing https://github.com/capnproto/go-capnproto2/issues/349; the idea is that the
method receiver should get a reference to this too to keep it alive for
the right amount of time.

### DIFF
--- a/internal/rc/rc.go
+++ b/internal/rc/rc.go
@@ -1,0 +1,30 @@
+package rc
+
+import "sync/atomic"
+
+type Releaser struct {
+	release  func()
+	refcount int32
+}
+
+func NewReleaser(init int32, release func()) *Releaser {
+	return &Releaser{
+		release:  release,
+		refcount: init,
+	}
+}
+
+func (rc *Releaser) Decr() {
+	newCount := atomic.AddInt32(&rc.refcount, -1)
+	if newCount == 0 {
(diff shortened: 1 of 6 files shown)


internal/rc/rc.go
```diff
@@ -0,0 +1,30 @@
...
+		refcount: init,
+	}
+}
+
+func (rc *Releaser) Decr() {
+	newCount := atomic.AddInt32(&rc.refcount, -1)
+	if newCount == 0 {
+		rc.release()
+		rc.release = nil
+	}
```
lthibault, line 22: Consider adding a test for a negative `newCount` and panicking if we over-release.

lthibault (changes requested):
I think this is good, but I'd like to take one last look at it after #327 & #373 has been merged, and after this has been rebased.

One small suggestion inline.

internal/rc/rc.go
zenhack: Done.

zenhack:
Merged #373 back into this.

lthibault (approved):

PR #379, opened 2022-12-13
Conn: wrap fields protected by mu in struct.

zenhack:
Per the comment, this makes it much easier to tell at a call site whether you need to be holding the lock. In the process of doing this, I also added comments on two methods where the need to hold the lock was not documented, and also pulled the `closed` channel out of this section -- it is not actually supposed to be protected by Conn.mu

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -86,11 +86,12 @@ type Conn struct {
 
 	sender mpsc.Queue[asyncSend]
 
-	mu sync.Mutex // mu protects all the following fields in lk.
-	// lk contains all the fields protected by mu; this makes it easy
-	// to tell at call sites whether you should or should not be holding
-	// the lock.
+	// lk contains all the fields that need to be protected by a mutex
+	// this makes it easy to tell at call sites whether you should or
+	// should not be holding the lock.
 	lk struct {
+		sync.Mutex // protects all the fields in lk.
+
 		closing bool // used to make shutdown() idempotent
 
 		// Tables
@@ -180,8 +181,8 @@ func NewConn(t Transport, opts *Options) *Conn {
 
 		c.er.ReportError(err) // ignores nil errors
 
-		c.mu.Lock()
-		defer c.mu.Unlock()
+		c.lk.Lock()
+		defer c.lk.Unlock()
 
 		if err = c.shutdown(err); err != nil {
 			c.er.ReportError(err)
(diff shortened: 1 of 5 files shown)


rpc/rpc.go
```diff
@@ -77,28 +77,32 @@ type Conn struct {
...
-	// Tables
-	questions  []*question
-	questionID idgen
-	answers    map[answerID]*answer
-	exports    []*expent
-	exportID   idgen
-	imports    map[importID]*impent
-	embargoes  []*embargo
-	embargoID  idgen
+	mu sync.Mutex // mu protects all the following fields in lk.
```
lthibault, line 89: Minor suggestion, maybe embed `sync.Mutex` directly into `lk`, and rename the latter to `tables`?

```go
lk struct {
	sync.Mutex

	closing bool // used to make shutdown() idempotent

	// Tables
	questions  []*question
	questionID idgen
	answers    map[answerID]*answer
	exports    []*expent
	exportID   idgen
	imports    map[importID]*impent
	embargoes  []*embargo
	embargoID  idgen
}
```

This would have the benefit of:

1. grouping the mutex with the fields it protects
2. correctly labelling the set of tables (it's not really a "lock", but a "table that needs locking")
3. being a bit clearer at the call-site, I think.

Regarding point 3, we would have calls that look like this:

```go
c.tables.Lock()
defer c.tables.Unlock()

c.tables.questions[0] // ...
```

Thoughts?

lthibault (changes requested):
Generally onboard with this strategy, but I find the API a bit awkward.  See suggestion inline.

rpc/rpc.go
zenhack: Not sure I'm a fan of `tables`; there are things in there other than tables. Also I shortened `locked` -> `lk` because it came up often enough that the verbosity was starting to get annoying...

I think embedding the mutex is a good idea though, I'll do that.

zenhack: (done)

zenhack: Hm, it's *just* closing that's extra though -- I wonder if it would be cleaner to:

- Actually define this as its own type, `tables`
- Delete `closing`
- Make the field a pointer
- Set it to nil when shutdown starts

...since we already nil out all these at the top of shutdown anyway.

I don't know that I want to do the functional changes in the same PR though. Thoughts?

zenhack: ...If we did this we might want to move the mutex back out of tables though, since there's otherwise no way to access it atomically...

lthibault: I actually think that makes sense, including the moving of the mutex back out of tables.  Also agree with doing this in a separate PR.

Is the current state something you can work with for the next few hours, or have we made a mess of things?  If it's the former, I say we merge it and revisit as soon as needed.

zenhack: This is I think a reasonable place to cut it; merge away.

PR #385, opened 2022-12-14
answer.destroy: avoid syncutil.Without

zenhack:
Instead, this reworks releaseList so that it's a list of arbitrary ReleaseFuncs, instead of just clients, and has destroy add msgReleaser.Decr to the releaseList it's already returning.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -1296,7 +1296,11 @@ func (c *Conn) recvPayload(payload rpccp.Payload) (_ capnp.Ptr, locals uintSet, 
 		var err error
 		mtab[i], err = c.recvCap(ptab.At(i))
 		if err != nil {
-			releaseList(mtab[:i]).release()
+			// FIXME: is it safe to release these while holding c.lk? In general,
+			// it's possible for Client.Release() to deadlock.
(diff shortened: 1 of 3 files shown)


rpc/rpc.go
```diff
@@ -1296,7 +1296,11 @@ func (c *Conn) recvPayload(payload rpccp.Payload) (_ capnp.Ptr, locals uintSet,
 		var err error
 		mtab[i], err = c.recvCap(ptab.At(i))
 		if err != nil {
-			releaseList(mtab[:i]).release()
+			// FIXME: is it safe to release these while holding c.lk? In general,
+			// it's possible for Client.Release() to deadlock.
```
zenhack, line 1300: Note: this was a problem with the previous implementation as well, I just noticed it while doing this. Perhaps we should open an issue to follow up?

lthibault: I'm not familiar enough to comment on the issue itself, but yes, let's definitely sort this one out.

lthibault (approved):

rpc/rpc.go
zenhack: Alright, opened an issue. Merging.

PR #390, opened 2022-12-15
Fix client leak in the server package

zenhack:
@lthibault found this in production; the test is distills the failure case.

The culprit seems to be a missing call to Promise.ReleaseClients() in the server package; this now mirrors the ReleaseFuncs returned by other implementations of ClientHook.

### DIFF
--- a/rpc/internal/testcapnp/test.capnp
+++ b/rpc/internal/testcapnp/test.capnp
@@ -18,3 +18,7 @@ interface CapArgsTest {
   call @0 (cap :Capability);
   self @1 () -> (self :CapArgsTest);
 }
+
+interface PingPongProvider {
+  pingPong @0 () -> (pingPong :PingPong);
+}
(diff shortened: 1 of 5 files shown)


zenhack:
@lthibault, FYI we have a custom test main that calls SetClientLeakFunc, so there's no need to set it manually in an individual test; it should detect leaks in any test automatically. (Just pushed another commit that fixes a data race that was caused by this).

zenhack:
That's a new one:

https://github.com/capnproto/go-capnproto2/actions/runs/3707839611/jobs/6284731111#step:5:1569

lthibault (dismissed):
LGTM.  Thanks for handling this one.

Not sure how you want to handle the test failure, so I'll leave the decision to merge up to you.

zenhack:
Pushed a fix for the "new" bug

lthibault (approved):

Issue #64, opened 2017-02-21
More convienent API for calling methods


zenhack:
The calling convention for clients is a bit funky, to say the least:

```go
remoteObj.Method(ctx, func(pkg.RemoteObj_method_Params p) error {
         p.SetFoo("hello")
         p.SetBar("world")
         return nil
})
```

Would be nice if we could do something more like:

```go
remoteObj.Method(ctx, "hello", "world")
```
I think the tradeoffs are similar to the pogs package, so maybe this would be a companion package to that. It would also be nice to be able use it with pogs without having to explicitly call Extract/Insert, but I'd want to find a way to avoid just making all of the args `interface{}`. Perhaps these wrappers could be generated along with the types from #63.

There's also a similar thing going on with the Server interfaces.

zombiezen:
How would you deal with default arguments?

zenhack:
We could just translate them mandatory arguments. They're a pretty rarely used feature in the sandstorm APIs, and not used at all in the schema that ship with capnproto itself. Given that, having to pass a sentinel value every now and then would fit with the go stdlib apis pretty seamlessly.

zombiezen:
Right, but that means that you can make a backwards-compatible change in your schema and break call sites.  That seems not ideal.  If you were to go down this route, I would think that you'd want to generate two methods (name bikeshedding aside):

```go
// This is the current Method
func (RemoteObjClient) Method(context.Context, func(RemoteObj_method_Params) error) RemoteObj_method_Results_Promise

// This is the newly generated Method which does not include any default arguments.
func (RemoteObjClient) MethodEasy(context.Context, foo string, bar string) RemoteObj_method_Results_Promise
```

I'm not convinced that there's an easier way to generate the server interfaces.

zenhack:
Quoting Ross Light (2017-03-15 18:52:42)

>    Right, but that means that you can make a backwards-compatible change
>    in your schema and break call sites. That seems not ideal.

It's worth noting we're already in a bad spot re: source compatibility.
Every time I update the sandstorm schema, something has grown new
methods and it breaks stuff.

Also worth nothing that fixing the call-site breakage would be very
mechanical (more so than the breakage we already deal with); you just
pass defaults for any missing arguments. One could even write a tool
to do this ala go fix.

> If you were to go down this route, I would think that you'd want to
> generate two methods (name bikeshedding aside):
>
> // This is the current Method
[...]

PR #391, opened 2022-12-16
Fix #382

### LINKED ISSUE #382: Incorrect locking in question.PipelineSend
Conn.sendMessage()'s comment states that the caller must be holding the lock, but question.PipelineSend invokes it in an explicit call to `syncutil.Without`. Unfortunately, just removing it deadlocks, since the nested call to newPipelineCallMessage tries to acquire the lock. This needs to be reworked so we're following all contracts.

zenhack:
This reworks the locking such that newPipelineCallMessage & newImportCallMessage expect the caller to already be holding the connection lock, and then removes the call to syncutil.Without in order to comply with sendMessage's contract.

There is a big downside here: it means that it is no longer OK for PlaceArgs to call into the RPC subsystem, as it could cause a deadlock.

We're planning on reworking this interface to get rid of PlaceArgs anyway (see #64), and this kind of code is not that common and generally easy to avoid, so in the interest of finishing out one batch of refactoring before starting the next, I am of the opinion that this is probably the right thing to do.

---

@lthibault, I want your opinion on the PlaceArgs thing.

### DIFF
--- a/rpc/import.go
+++ b/rpc/import.go
@@ -98,27 +98,25 @@ func (ic *importClient) Send(ctx context.Context, s capnp.Send) (*capnp.Answer, 
 	q := ic.c.newQuestion(s.Method)
 
 	// Send call message.
-	syncutil.Without(&ic.c.lk, func() {
-		ic.c.sendMessage(ctx, func(m rpccp.Message) error {
-			return ic.c.newImportCallMessage(m, ic.id, q.id, s)
-		}, func(err error) {
-			if err != nil {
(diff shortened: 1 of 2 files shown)


zenhack:
Failure looks like the same as in https://github.com/capnproto/go-capnproto2/pull/376#issuecomment-1347568475

lthibault (approved):
LGTM.  The limitation on `PlaceArgs` sounds sensible.  I'm not making any RPC calls inside of that callback in my code, and I suspect few people are going to try it (famous last words...).

PR #392, opened 2022-12-16
Refactor use of releaseList

zenhack:
This patch:

- Moves releaseList out of exports.go into its own file
- Uppercases .Release(); per the commit message there's an interface begging to be defined here.
- Give releaseList an .Add() method
- Changes methods that return releaseLists to accept pointers to them as arguments instead, and use .Add()
- Where possible, centralize creation & release of releaseList (see handleBootstrap()). This has the nice effect of killing a couple more uses of syncutil.Without.

There are a few more places where this could help clean things up, but the needed changes are a bit less mechanical so I left them as future work.

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -194,8 +194,7 @@ func (ans *answer) Return(e error) {
 		return
 	}
 	rl := &releaseList{}
-	err := ans.sendReturn(rl)
-	if err != nil {
+	if err := ans.sendReturn(rl); err != nil {
 		select {
 		case <-ans.c.bgctx.Done():
 		default:
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -185,14 +185,16 @@ func (ans *answer) setBootstrap(c capnp.Client) error {
 func (ans *answer) Return(e error) {
 	ans.c.lk.Lock()
 	if e != nil {
-		rl := ans.sendException(e)
+		rl := &releaseList{}
+		ans.sendException(rl, e)
 		ans.c.lk.Unlock()
 		rl.Release()
 		ans.pcalls.Wait()
 		ans.c.tasks.Done() // added by handleCall
 		return
 	}
-	rl, err := ans.sendReturn()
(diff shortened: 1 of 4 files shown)


rpc/answer.go
```diff
@@ -185,14 +185,16 @@ func (ans *answer) setBootstrap(c capnp.Client) error {
...
 		ans.c.lk.Unlock()
-		rl.release()
+		rl.Release()
 		ans.pcalls.Wait()
 		ans.c.tasks.Done() // added by handleCall
 		return
 	}
-	rl, err := ans.sendReturn()
+	rl := &releaseList{}
+	err := ans.sendReturn(rl)
```
lthibault, line 197: Nit:  this could be inlined with the `if` statement below.  It's marginally cleaner that way, IMO.

rpc/releaselist.go
```diff
@@ -0,0 +1,18 @@
...
+import "capnproto.org/go/capnp/v3"
+
+type releaseList []capnp.ReleaseFunc
+
+func (rl releaseList) Release() {
+	for _, r := range rl {
+		r()
+	}
+	for i := range rl {
+		rl[i] = func() {}
```
lthibault, line 12: Nit:  I'm not sure what the compiler/runtime implications are of assigning nop functions, but I suspect it might allocate a new anonymous function pointer each time.  It's not a huge deal, but since we're already tweaking `releaseList`, maybe we should nil these out and add a nil check in the previous loop instead?

lthibault (dismissed):
LGTM.  See inline for two minor suggestions, which you can feel free to ignore.

rpc/releaselist.go
zenhack: I think it's more readable the way you suggest, so done (and I also merged the two loops) -- but functions can only be compared to nil, so  I'm like 99.9% sure this won't allocate; it should be able to just assign the same function pointer each time.

lthibault (approved):

Issue #377, opened 2022-12-13
Failure in TestRecvDisembargo


zenhack:
From https://github.com/capnproto/go-capnproto2/actions/runs/3662127729/jobs/6190989469#step:5:1530

```
 === RUN   TestRecvDisembargo
Error:     level0_test.go:2085: conn error: rpc: incoming disembargo: answer ID 55 has not sent return
Error:     level1_test.go:767: don't know how to handle abort; skipping
Error:     level1_test.go:674: recvMessage(ctx, p2): transport: stream transport: receive: io: read/write on closed pipe
--- FAIL: TestRecvDisembargo (0.17s)
```

lthibault:
Closing as no longer reproducible on current main. TestRecvDisembargo was run together with the other historical RPC regression tests 50 times without failure. We can reopen if the failure recurs with a current revision and reproducible trace.

PR #395, opened 2022-12-20
Clean up/fix locking in handleReturn

zenhack:
This cleans up/fixes some of the locking towards the end of handleReturn; previously this was super convoluted, in part because we'd just put the whole rest of the function in a giant `go func() {...}()` at some point to solve a deadlock without cleaning it up in a more structural way; see the now-deleted `TODO(cleanup)`.

This also adds a few comments related to the reasoning behind some conditionals, which are a bit obtuse. I think this probably could be cleaned up further, but this maintains the existing behavior while documenting it a bit better at least.

Finally, while doing the above I noticed that sendMessage() was being called without holding c.lk, so I added a call to syncutil.With to fix that.

After this patch there are only three remaining calls to `syncutil.Without` in the rpc package.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -1000,39 +1000,41 @@ func (c *Conn) handleReturn(ctx context.Context, ret rpccp.Return, release capnp
 			release()
 		}
 
-		// Send disembargoes.  Failing to send one of these just never lifts
-		// the embargo on our side, but doesn't cause a leak.
-		//
-		// TODO(soon): make embargo resolve to error client.
-		for _, s := range pr.disembargoes {
-			c.sendMessage(ctx, s.buildDisembargo, func(err error) {
+		syncutil.With(&c.lk, func() {
+			// Send disembargoes.  Failing to send one of these just never lifts
+			// the embargo on our side, but doesn't cause a leak.
+			//
+			// TODO(soon): make embargo resolve to error client.
+			for _, s := range pr.disembargoes {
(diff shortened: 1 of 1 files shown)


rpc/rpc.go
```diff
@@ -978,71 +978,63 @@ func (c *Conn) handleReturn(ctx context.Context, ret rpccp.Return, release capnp
...
-		case q.bootstrapPromise != nil:
-			syncutil.Without(&c.lk, func() {
-				q.p.Resolve(pr.result, pr.err)
-				q.bootstrapPromise.Fulfill(q.p.Answer().Client())
-				q.p.ReleaseClients()
-				release()
-			})
-		case pr.err != nil:
-			// TODO(someday): send unimplemented message back to remote if
-			// pr.unimplemented == true.
```
zenhack, line 1000: Note, I just deleted this comment; I'm 99.99% sure this is either wrong or just unclear: the unimplemented message is not for when clients return unimplemented exceptions; those should should just do a normal error return with type = unimplemented.

zenhack:
You might want to look at the individual commits instead of the whole diff, which is horrible. The latter commit indented a big block of code which git never seems to understand very well.

zenhack:
Source of the failure isn't immediately obvious to me; I'll stare at it in the morning (though I would hazard a guess it's been there and is not related to this patch).

lthibault (approved):

zenhack:
Last failure looks like #377, so not the same bug. Once we've nailed down these more structural improvements, we should probably do another pass of "get CI to stop failing"

PR #396, opened 2022-12-21
Clean up/fix locking in handleDisembargo

zenhack:
- Avoid syncutil.Without
- Hold c.lk when invoking sendMessage().

The latter requires us to move some of the release()es later, but this is fine.

PR #397, opened 2022-12-21
handleBootstrap: avoid a use of syncutil.Without()

zenhack:
Stacked on top of #396; merge that first.

Should be no functional change here.

Only 2 more call sites!


lthibault (approved):

lthibault:
```
Error:         level0_test.go:2124: conn error: rpc: send release: send message: context canceled
```

Seems like this might be fixed by #396 ?

PR #398, opened 2022-12-26
Clean up locking in Conn.shutdown()

zenhack:
This patch changes the contract of shutdown() such that it now expects the caller *not* to be holding c.lk. This allows it to acquire and release the lock itself as needed, thereby avoiding the need for syncutil.Without or just plain unlock/lock pairs.

This also reworks its subroutines to avoid unlock/lock pairs. The result is a lot easier to follow imo.

---

There is now only one use of syncutil.Without left in the codebase, though there are likely several other logical unlock/lock pairs to be rooted out.

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -199,11 +199,12 @@ func (ans *answer) Return(e error) {
 		case <-ans.c.bgctx.Done():
 		default:
 			ans.c.tasks.Done() // added by handleCall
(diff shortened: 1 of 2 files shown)


zenhack:
Failure looks like #394, re-running.

lthibault (approved):
I agree this is much cleaner 👍

PR #399, opened 2022-12-26
Simplify rpc/answer.Return()

zenhack:
This simplifies the implementation of answer.Return(). stacked on top of #398; review and merge that first.

This patch:

- Gets rid of the unnecessary select; shutdown is idempotent, so it's fine to do it unconditionally; we don't need to check bgctx.
- Reworks the locking so that we're using synctuil.With; makes it much easier to follow.
- Moves ans.pcalls.Wait() to a defer, rather than replicating it at each return site (unfortunately we can't do the same for tasks.Done(), because it needs to be called before invoking shutdown()


lthibault (approved):
👍

PR #401, opened 2022-12-27
Tweak locking around Conn.sender

zenhack:
Contrary to the comment, there was no actual need to not be holding the lock when calling answer.sendMsg, and flipping the contract allows us to remove a superfluous unlock/lock pair.

This also changes the send queue back to using `spsc` and protects it with Conn.lk, since we no longer add to the send queue when we're not holding the connection lock anyway. We'd tried this in the past, but backed it out due to the now-removed unlock/lock pair.

Doing this required splitting it into two separate fields, so the receive end could still be outside Conn.lk.

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -116,7 +116,7 @@ func (c *Conn) newReturn() (_ rpccp.Return, sendMsg func(), _ *rc.Releaser, _ er
 	releaser := rc.NewReleaser(2, outMsg.Release)
 
 	return ret, func() {
-		c.sender.Send(asyncSend{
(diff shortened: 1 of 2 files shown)


zenhack:
Tacked on one more fix; see the commit message on the last commit.

lthibault (approved):
Looks good.  I really like the way this lock-reduction effort is panning out overall 👍 

Quick note:  I'm going to let you hit the merge button until we get around to fixing CI.  I know most of these failures are known/probably-harmless issues, but I'd rather leave the final decision to merge to you.

PR #403, opened 2022-12-27
handleCall: cleanups & locking simplifications

zenhack:
This gets us to a place where locking in handleCall is strictly hierarchical; there is one call to syncutil.With and then later a `lock(); defer unlock()`.

I don't love the way we're using releaseList to do the actual Recv calls, but factoring those out into one codepath turned out to be really thorny. This is at least cleaner than what we had before, and gets this into a shape where it can be easily adapted to the withLocked() design we'd discussed.

I'm also not a fan of how many codepaths have to call releaseCall, but cleaning up resource reclamation is another broader topic on my list...

This also includes a couple minor cleanups in 561756942fa69284bd6009cbfce5ff2def38d425 and 28f46dece574b1e6382f92ed99dd416e5a62a240

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -663,6 +663,16 @@ func (c *Conn) handleBootstrap(ctx context.Context, id answerID) error {
 	return err
 }
 
+func makeIdempotent(f func()) func() {
+	called := false
+	return func() {
+		if !called {
+			called = true
+			f()
+		}
+	}
+}
+
 func (c *Conn) handleCall(ctx context.Context, call rpccp.Call, releaseCall capnp.ReleaseFunc) error {
 	rl := &releaseList{}
 	defer rl.Release()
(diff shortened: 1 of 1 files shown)


rpc/rpc.go
```diff
@@ -663,6 +663,16 @@ func (c *Conn) handleBootstrap(ctx context.Context, id answerID) error {
 	return err
 }
 
+func makeIdempotent(f func()) func() {
```
lthibault, line 666: This is a minor issue of taste/style, but I find that `idempotent(foo)` somehow reads better.   I recognize it runs somewhat contrary to the verb-y function names commonly used in Go, so it's your call.

lthibault (approved):

rpc/rpc.go
zenhack: Done.

lthibault (approved):

PR #405, opened 2022-12-28
Add handler for unknown method

hspaay:
This is a proposed solution to be able to customize the handling of unknown methods. The default behavior is unchanged but a hook allows customized behavior.

The hook can be used to dynamically create a method and have it invoked as if it existed all along.
I was able to use this in a proxy service that dynamically obtains capabilities/methods from other services and forwards requests to those services.

### DIFF
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -111,7 +111,7 @@ func TestServerCall(t *testing.T) {
 			sm.Impl = func(ctx context.Context, call *server.Call) error {
 				echoArgs := air.Echo_echo_Params(call.Args())
 				inText, err := echoArgs.In()
-				assert.NoError(t, err)
+				require.NoError(t, err)
 				proxyReceived.Store(inText)
 				// pretend we received an answer
 				echo := air.Echo_echo{Call: call}
--- a/server/server_test.go
+++ b/server/server_test.go
@@ -3,12 +3,17 @@ package server_test
 import (
 	"context"
 	"errors"
+	"net"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 
 	"capnproto.org/go/capnp/v3"
 	air "capnproto.org/go/capnp/v3/internal/aircraftlib"
+	"capnproto.org/go/capnp/v3/rpc"
 	"capnproto.org/go/capnp/v3/server"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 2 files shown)


hspaay:
This is as barebones as it gets. Lets start picking at it. Interface? yes/no/good enough anyone? :)

lthibault (approved):
LGTM 👍

zenhack:
Could use a test, but I'm fine with this. Before we tag 3.0 I think I want to do the interface reworking we talked about though.

lthibault:
Ah yeah, +1 to the test.

Also, @zenhack, can you open an issue for the interface refactor?  I think you're in the best position to accurately spec it out.

server/server_test.go
```diff
@@ -91,6 +96,63 @@ func TestServerCall(t *testing.T) {
 			}
 		}
 	})
+	t.Run("Unimplemented hook", func(t *testing.T) {
```
lthibault, line 99: Just FYI, the testing framework will replace whitespaces with underscores in test names.  This doesn't actually break anything -- just thought I'd mention it.

lthibault (changes requested):

lthibault:
I've flagged some minor things, but this basically LGTM.  I think we're good to merge after these fixes.

server/server_test.go
```diff
@@ -91,6 +96,63 @@ func TestServerCall(t *testing.T) {
...
+		srv.HandleUnknownMethod = func(method capnp.Method) *server.Method {
+			sm := server.Method{
+				Method: method,
+				Impl:   nil,
+			}
+			sm.Impl = func(ctx context.Context, call *server.Call) error {
+				t.Log("welcome to blank")
+				echoArgs := air.Echo_echo_Params(call.Args())
+				inText, err := echoArgs.In()
+				assert.NoError(t, err)
```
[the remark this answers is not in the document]
hspaay: The motivation for this is that assert in this place helps pinpoint the problem early without having to chase the rabbit down the hole.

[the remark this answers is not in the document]
lthibault: Fair enough.  Just be aware that it won't actually abort execution of that function (which is likely to cause problems that make the output _more_ confusing).

If you want something that hard-stops on an error, use `require.Foo(...)`.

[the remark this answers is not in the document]
hspaay: Good point. Changed it into require. Thanks :)

lthibault (dismissed):
LGTM.

server/server_test.go
```diff
@@ -91,6 +96,60 @@ func TestServerCall(t *testing.T) {
...
+				// pretend we received an answer
+				echo := air.Echo_echo{Call: call}
+				resp, _ := echo.AllocResults()
+				err = resp.SetOut(inText)
+				return err
+			}
+			return &sm
+		}
+		blankBoot := capnp.NewClient(srv)
+		lis, err := net.Listen("tcp", ":0")
```
zenhack, line 125: You can simplify this and entirely skip setting up a connection with something on the other side and just use `blankBoot` in place of `blankClient`

zenhack (changes requested):
One minor thing

server/server_test.go
hspaay: @zenhack wouldn't that invoke 'Send' instead of 'Recv'?  The hook was intended for acting as a proxy forwarder and it doesn't exist in Send.

zenhack: O.O can't believe I missed this on my first look: Yeah, it would, and you really don't want the two to have different behavior -- the only difference that's supposed to be there is who allocates the arguments. Probably we should move the logic for method lookup into Server.start so it's not duplicated in both Send and Recv.

hspaay: Indeed it looks like there is some room for cleanup here as well. Combine methods.find, although that might not be as easy as it sounds.
{rant} Looking at the Send() function the return statement is rather confusing. I tried reading it a few times and am still not quite sure what is happening in what order. I've been bitten in the past by statements that do more  than one thing. In this case the return statement of Send() is akin to a rattlesnake den. The casual reader will benefit from having it split into a statement for each of the things it is doing. It looks like it needs the result of methods.find before srv.start is called but this is not quite clear to me. 

Would it be okay if  I add the hook to Send()  in the same way it is added to Recv and leave the cleanup to more capable folks?

zenhack:
Quoting Henk (2023-01-03 00:10:07)

> The casual reader will benefit from having it split into a statement
> for each of the things it is doing.

Yeah, I agree; it would be better to split it up into multiple
statements.

> It looks like it needs the result of methods.find before srv.start is
> called but this is not quite clear to me.

Yeah, I think that's fixable without too much major surgery, but

> Would it be okay if I add the hook to Send() in the same way it is
> added to Recv and leave the cleanup to more capable folks?

I'd be ok with that; I won't force you to shave that yak.

hspaay:
> I'd be ok with that; I won't force you to shave that yak.

Thank you.

zenhack:
I'd still like to remove the connection from that test, but other than that LGTM

zenhack (approved):

PR #408, opened 2022-12-29
Improve type safety wrt locking.

zenhack:
This patch introduces a lockedConn type:
    
```
type lockedConn Conn
```
    
...and moves all methods that require the caller to hold c.lk onto it. It provides methods for running functions while holding the lock.
 
This makes it much harder to introduce bugs where a method is being called without holding the lock as required, or vice versa.
    
The patch also fixes a few call sites that were in fact not doing locking correctly, which were caught by the type checker.

The second commit also cleans up some bitrot in the implementation advice comment in rpc.go

### DIFF
--- a/rpc/answer.go
+++ b/rpc/answer.go
@@ -198,8 +198,8 @@ func (ans *answer) Return(e error) {
 
 	var err error
 
-	syncutil.With(&ans.c.lk, func() {
-		err = ans.sendReturn(rl)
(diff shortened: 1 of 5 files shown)


lthibault (approved):
LGTM.

>The patch also fixes a few call sites that were in fact not doing locking correctly, which were caught by the type checker.

Paying dividends already 😃

Issue #364, opened 2022-12-04
Exorcise fmt from the library packages


zenhack:
We discussed this a while back on matrix: I'm doing some stuff where I'm compiling Go to webassembly and using capnp to communicate between a browser and a server. I'm using TinyGo to try to keep executable sizes down, but the `fmt` package in particular pulls in a lot of stuff. There's no good reason why we should need to pull it in as a dependency; I'd like to avoid its use in the library parts of this (tests and code generator are fine) so we can output smaller binaries where it matters.

PR #412, opened 2023-01-02
Get rid of formatting helpers that call out to fmt

zenhack:
e.g. Failedf. More work towards #364.

This mostly leaves direct calls to the fmt package.


zenhack:
Tacked on another commit that gets rid of  direct calls in rpc & transport.

lthibault (approved):

PR #418, opened 2023-01-09
Fix occasional spurious test failure.

zenhack:
Along the lines of 3fc50651bf0b99184418bc046b922d82b19edddc, just with a different test: occasionally, with the race detector and other tests running in parallel, I will get a timeout here. But I think that's just that there's no reason in principle this can't take more than two seconds, so under extreme circumstances it may. So this patch removes the timeout; if there's a real bug the channel receive will just hang and eventually the test-suite-wide timeout will catch it.

### DIFF
--- a/rpc/level0_test.go
+++ b/rpc/level0_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"math"
 	"net"
 	"os"
 	"runtime"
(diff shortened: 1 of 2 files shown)


zenhack:
...tacked on a second commit fixing a similar problem in another test.

Note that the first of these I have seen in CI a few times.

lthibault (approved):
LGTM.

lthibault:
```
=== RUN   TestPipelineCall
Error:     server_test.go:272: GetNumber() #1 = 1; want 0
Error:     server_test.go:278: GetNumber() #2 = 0; want 1
```

lthibault:
I got the tests to pass, so I went ahead and merged.

zenhack:
Yeah, that is new -- pretty confident this isn't the source, but we should dig into that.

lthibault:
I've opened https://github.com/capnproto/go-capnproto2/issues/420 to track it.

PR #422, opened 2023-01-12
Don't report nil error when shutting down import client.

lthibault:
Adds a nil error check before reporting an error in `importClient.Shutdown()`.


lthibault:
I ended up finding a few more places where we were calling `Annotate` with a potentially-nil error.  See commits.

zenhack (approved):

PR #424, opened 2023-01-13
Fix #420

### LINKED ISSUE #420: TestPipelineCall fails
```
=== RUN   TestPipelineCall
Error:     server_test.go:272: GetNumber() #1 = 1; want 0
Error:     server_test.go:278: GetNumber() #2 = 0; want 1
```

This is new, isn't it? ☝️

_Originally posted by @lthibault in https://github.com/capnproto/go-capnproto2/issues/418#issuecomment-1375740560_

zenhack:
This fixes #420 (without re-introducing #394, or #349). Per the last commit:

> 420 was a regression introduced by our fix for 394; if we Return()
before we fulfill(), this can result in out of order message delivery,
since incoming calls can be made directly on clients in the result
before all queued calls are drained.
>
> Previously, if we fulfilled() before Return(), we would get a data race
as fillPayloadCapTable modified the message while pipelined calls read
it. Having split Return(error) into PrepareReturn(error) and Return(),
it is now safe to move just the latter after the call to fulfill().

I am now unable to reproduce any of the referenced issues.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -882,7 +882,9 @@ type Returner interface {
 	AllocResults(sz ObjectSize) (Struct, error)
 
 	// PrepareReturn finalizes the return message. The method call will
-	// resolve successfully if e is nil, or otherwise it will fail.
+	// resolve successfully if e is nil, or otherwise it will return an
+	// exception to the caller.
+	//
 	// PrepareReturn must be called once.
 	//
 	// After PrepareReturn is invoked, no goroutine may modify the message
--- a/capability.go
+++ b/capability.go
@@ -857,7 +857,8 @@ func (r Recv) AllocResults(sz ObjectSize) (Struct, error) {
 // Return ends the method call successfully, releasing the arguments.
 func (r Recv) Return() {
 	r.ReleaseArgs()
-	r.Returner.Return(nil)
+	r.Returner.PrepareReturn(nil)
+	r.Returner.Return()
 }
 
 // Reject ends the method call with an error, releasing the arguments.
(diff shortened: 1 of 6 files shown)


capability.go
```diff
@@ -879,13 +881,21 @@ type Returner interface {
 	// ReleaseResults is called.
 	AllocResults(sz ObjectSize) (Struct, error)
 
-	// Return resolves the method call successfully if e is nil, or failure
-	// otherwise.  Return must be called once.
+	// PrepareReturn finalizes the return message. The method call will
+	// resolve successfully if e is nil, or otherwise it will fail.
```
lthibault, line 885: Nit:  I initially misunderstood this to mean "the actual sending of the return message will fail".  Consider updating docstring to something like:

>The method call will resolve successfully if e is nil.  Otherwise, it will return an exception to the caller.

lthibault (approved):

lthibault (approved):

lthibault:
```
Error:         level0_test.go:2126: conn error: rpc: send release: send message: context canceled
```

zenhack:
...that's one of the existing ones.

lthibault:
Yeah, I thought so, but couldn't find an issue for it.  The error in https://github.com/capnproto/go-capnproto2/issues/377 is a different one.

Should I open an issue for this?  (Re-running CI in the meantime.)

zenhack:
Yeah, why don't you open a separate one; I think this one's been around a while but I keep binning it with #377

PR #427, opened 2023-01-13
Remove extra msgReleaser.Decr()

zenhack:
Noticed this while working on #424.  Per the comment below the if statement, we don't actually abort sending the message in this case, so doing a release here makes no sense. Probably I wasn't paying close enough attention when I added this.


lthibault (approved):

zenhack:
Hm, the CI interface won't actually show me the error message...

lthibault:
Yeah it's been doing that from time to time.  I re-ran the tests after seeing the same `TestSendDisembargo` error the first time.  I'm guessing it's the same one again.

Issue #423, opened 2023-01-13
Dropping the last reference to a Server should not cancel outstanding calls


zenhack:
Right now, the server package invokes method calls with a context that is canceled when *either* the server's context is cancelled (i.e. the last reference was dropped), or the context passed in the original call is canceled.

I think this is not the semantics we want; it means you can't write code like:

```
fut, rel := cap.Foo(ctx, ...)
defer rel()
cap.Release() // Done calling methods on cap, drop the reference now.
res, err := fut.Struct() // may spuriously return an error, since we released cap
```

Given that the caller already has a way to cancel the method call (cancel the context that was passed in), I think we should change this so that that is the only way the call will be cancelled. Note that in the case where the call is coming in from a connection, the call's context will still be bound to the lifetime of the connection.

I think I'm hitting this in tempest in a case where I have pipelined streaming calls on a cap, and then I .Release() it after sending the last call, inadvertently causing in-flight calls to be cancelled.

@lthibault, thoughts?

lthibault:
I actually have some rather important code that depends on existing semantics.  Specifically, long-running method calls (forked via `call.Go()`) needs a mechanism for detecting that the server is being shut down, so that they can return.  Otherwise, the `Server.Shutdown()` method will block indefinitely.

Here's a potential solution though.  We could take a page out of `net/http`'s book, and add a `.Context()` method to the RPC's `call` parameter.  The idea is that the `ctx` explicitly passed into `Foo(ctx, call)` corresponds to the _server_ context,  and `call.Context()` corresponds to the _call's_ context.  In your example, I think you would only want `fut.Struct()` to return the _call_ context's error, as it would indicate a business-logic error in the server's method.  The caller is already aware, as you've noted, of any cancelation signal to the server context.

In many ways, I think this would give us the best of both worlds.  We would no longer have to fork a goroutine in `handleCalls` to bind the call and server contexts, and it can be useful to distinguish between "call is aborted" and "server is shutting down".

zenhack:
I have a visceral reaction to just exposing two separate contexts to app code. I'll think on it.

lthibault:
Yeah, I had the same feeling at first.  If I may anticipate your concerns, the thing that initially bothered me was the issue of deciding which context to pass to any subroutines in the RPC method.  For example:

```go
func (s someServer) Foo(ctx context.Context, call Fooer_foo) error {
    // ...

[...]

PR #428, opened 2023-01-14
Server.handleCalls: deal with ack/Go correctly when shutting down.

zenhack:
Previously, the logic to check for an ack after returning from handleCall was missing from the loop that drains the queue after the server's context is canceled. This means that if when draining the queue, if a method handle calls .Go() before checking the context, there will be two goroutines servicing method calls when it returns.

This patch folds the shutdown into the same loop, so that the logic is the same, and so that the shutdown logic respects acks.

Since the queue is meant to be single-consumer, the exact consequences of this are a bit hard to reason about, though I have noticed that after fixing this some of the behavior I had attributed to #423 is less common.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -171,7 +171,14 @@ func (srv *Server) handleCalls(ctx context.Context) {
 	for {
 		call, err := srv.callQueue.Recv(ctx)
 		if err != nil {
-			break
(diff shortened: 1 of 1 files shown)


lthibault (approved):

lthibault:
Oh look!  Successful CI 🎉

Issue #426, opened 2023-01-13
Failure in TestSendDisembargo/SendQueuedResultToCaller


lthibault:
```
--- FAIL: TestSendDisembargo (0.01s)
    --- FAIL: TestSendDisembargo/SendQueuedResultToCaller (0.01s)
Error:         level1_test.go:285: Received finish message; want disembargo
Error:         level0_test.go:2126: conn error: rpc: send release: send message: context canceled
```

_Originally posted by @lthibault in https://github.com/capnproto/go-capnproto2/issues/424#issuecomment-1382569423_

PR #429, opened 2023-01-14
Add inflight message limiter

lthibault:
Adds a `FlowController` that counts the number of inflight messages, without regards for message size.
Bundles minor improvements to `TestFixed` as well in a separate commit.


lthibault:
#426. Re-running CI.

lthibault:
Closing per discussion on Matrix.  @zenhack made the point that we should not increase the API surface in response to a hacky implementation.  See [here](https://github.com/capnproto/go-capnproto2/blob/993783d252bc4843fcb85fecb11df3c4aceb2636/capability.go#L345-L358).

PR #430, opened 2023-01-15
Panic on flowcontrol deadlock

lthibault:
Fixes issue where `fixedLimiter` would silently deadlock when the size of a message exceeds the maximum total reservations.

Minor code cleanup included in separate commits.

### DIFF
--- a/flowcontrol/fixed.go
+++ b/flowcontrol/fixed.go
@@ -26,8 +26,7 @@ func (fl *fixedLimiter) StartMessage(ctx context.Context, size uint64) (gotRespo
 	//        reservation on the semaphore. We can't return an error because it
 	//        is currently ignored by the caller.
 	if int64(size) > fl.size {
-		const s = "StartMessage(): message size %d is too large (max %d)"
-		panic(fmt.Sprintf(s, size, fl.size))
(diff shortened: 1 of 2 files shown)


flowcontrol/fixed.go
```diff
@@ -3,35 +3,38 @@ package flowcontrol
...
-		return nil, fmt.Errorf(
-			"StartMessage(): limit %v is too large (max %v)",
-			size, int64(math.MaxInt64),
-		)
+	// HACK:  avoid dead-locking if the size of the message exceeds the maximum
+	//        reservation on the semaphore. We can't return an error because it
+	//        is currently ignored by the caller.
+	if int64(size) > fl.size {
+		const s = "StartMessage(): message size %d is too large (max %d)"
+		panic(fmt.Sprintf(s, size, fl.size))
```
zenhack, line 30: IMO factoring the format string out into a constant like this just makes things more confusing. There's also probably a linter out there that will yell at you for this...

zenhack (changes requested):

flowcontrol/fixed.go
lthibault: Done

zenhack (approved):

PR #431, opened 2023-01-15
Add context to 'too many segments' error in decoder.

lthibault:
Adds maximum observed segment ID and max segments for the given stream to error returned by decoder.

### DIFF
--- a/message.go
+++ b/message.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"encoding/binary"
 	"errors"
+	"fmt"
 	"io"
 	"sync"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -688,7 +689,8 @@ func (d *Decoder) Decode() (*Message, error) {
 	}
 	maxSeg := SegmentID(binary.LittleEndian.Uint32(d.wordbuf[:]))
 	if maxSeg > maxStreamSegments {
-		return nil, errors.New("decode: too many segments to decode")
+		return nil, fmt.Errorf("decode: too many segments (%d) to decode (max=%d)",
```
zenhack, line 692: Per #364, can we void fmt here and use strconv instead?

lthibault: Sure.  Should I define a separate error type?

zenhack: You don't need to; you could also just concatenate strings and keep using errors.New. Your call.

zenhack: (You probably want to use internal/str.Utod instead of strconv directly)

zenhack (approved):

PR #432, opened 2023-01-15
Add Release() method to capnp.Arena.

lthibault:
Exports the `Release()` method from the public `Arena` interface, to support pooling of arenas.  Adds `Release()` method to `SingleSegmentArena`, and other private arena types.

### DIFF
--- a/message.go
+++ b/message.go
@@ -407,13 +407,11 @@ type Arena interface {
 	// returned byte slice.
 	Allocate(minsz Size, segs map[SegmentID]*Segment) (SegmentID, []byte, error)
 
-	// Return this arena to an internal sync.Pool of arenas that can be
-	// re-used, which can help reduce memory allocations.
-	//
-	// All segments will be zeroed before re-use.
+	// Release all resources associated with the Arena, allowing it to
+	// be reused in another message.
 	//
 	// Calling Release is optional; if not done, the garbage collector
-	// will release the memory per usual.
+	// will release resources per usual.
 	Release()
 }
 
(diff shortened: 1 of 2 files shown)


message.go
```diff
@@ -406,6 +406,15 @@ type Arena interface {
 	// arena is responsible for preserving the existing data in the
 	// returned byte slice.
 	Allocate(minsz Size, segs map[SegmentID]*Segment) (SegmentID, []byte, error)
+
+	// Return this arena to an internal sync.Pool of arenas that can be
+	// re-used, which can help reduce memory allocations.
```
zenhack, line 411: I'd kindof like to keep this a little more abstract, and find a way to  express this without reference to the exact implementation. In particular, you've defined this as a no-op on roSingleSegment and testArena, which seems fine but is inconsistent with the docs.

zenhack (changes requested):
Minor gripe about the docs, but otherwise this looks ok.

message.go
lthibault: Ah, that's a good thought.  Fixed.

zenhack: Hm, I worry the new version is a little misleading and suggests something more like msg.Reset(); the caller should not directly re-use the arena.

lthibault: Baah, you're right.  In fact, I got myself mixed up, as you can see.

zenhack (approved):

lthibault:
```
Error:         level0_test.go:2126: conn error: rpc: send release: send message: context canceled`
```

This again.  Merging.

PR #436, opened 2023-01-22
CI: only run the tests repeatedly, not compile them.

zenhack:
This patch tweaks CI so that the phase that runs the rpc tests repeatedly still only builds them once -- this should speed things up with no real downside.


zenhack:
Looks like #426. I wish I could reproduce it this reliably locally.

zenhack:
If CI actually passes with this change I want to re-run it a few more times to make sure it still triggers our outstanding bugs; if not I may want to hold off on this change to avoid silencing those.

PR #437, opened 2023-01-22
CI: produce a trace file.

zenhack:
This patch adjusts CI to generate a trace when running the rpc tests repeatedly, and to make that trace available for download.


zenhack:
If CI actually passes with this change I want to re-run it a few more times to make sure it still triggers our outstanding bugs; if not I may want to hold off on this change to avoid silencing those.

zenhack:
The last run reproduced it, but it looks like I got the path wrong for the file or something? needs more investigation.

PR #438, opened 2023-01-28
Define .cancelTasks() on lockedConn instead of Conn.

zenhack:
Similar to the recent patch involving findEmbargo


zenhack:
Failure is #426

lthibault (approved):

PR #440, opened 2023-01-31
answer.sendMsg(): take a lockedConn

zenhack:
This should make this API a bit harder to misuse; did this in the process of sanity checking we were doing locking correctly (we were). No functional change.


zenhack:
#426 again.

Issue #63, opened 2017-02-19
pogs: codegen


zenhack:
As mentioned in #62, I'm working on table-based tests for go.sandstorm. For this to work I need pogs-style structs for basically every type in the sandstorm API. It goes without saying that doing this manually is rather tedious; it would be nice if go-capnpc could generate types usable with pogs.

zombiezen:
Interesting. I suppose you could have an alternate code generator program that dumps out the structs you want, but I think it would be hard to have the transitive closure of types in general.

zenhack:
Quoting Ross Light (2017-02-21 11:18:32)

> I think it would be hard to have the transitive closure of types in
> general.

Not sure I see why?

zombiezen:
Would you maintain a parallel tree of the imports for referenced pogs types, or would you inline all of them?  Both have tradeoffs, but if you're using the tool in a semi-automated way, either could work.

zenhack:
I had been thinking the former.

zombiezen:
Either approach is fine.  One thing to note: the way I use pogs will align closely (but not exactly) with an underlying schema, so some amount of manual intervention is to be expected.  For an example, see https://github.com/zombiezen/mcm/blob/master/internal/catpogs/catpogs.go

zenhack:
Just wanted to record a problem that's been in my head re: a problem that crops up with this and default arguments: suppose a schema adds a new field to a struct, which has a custom default argument. Updating the output of the code generator will silently change the value created by client code, as adding the field means we'd be setting the value to Go's zero value, rather than the default value which would have been effectively generated by the code that just didn't write the value.

hspaay:
Below my 2c's from a newbie that is struggling with the steep learning curve. If it wasn't for the excellent help I received I would have thrown in the towel already and continued with the non-capability gRPC. That would have been a real shame :smile: 
In short, I think that the adoption of go-capnproto would be greatly helped by a higher level API that can work with plain-old-go-objects.

One example, currently you cannot easily use the generated types elsewhere in the code as they are tied to the RPC messaging. Few people would be willing to spread all the rpc messaging aspects in their code that is not doing any messaging. I certainly won't. The result is that the Go implementation of an app now has to add a duplicate type definition, one for plain Go and the other for the capnp-rpc. Ouch.

By far the most ideal solution is that the generated code can use plain old Go types. Maybe as a separate layer on top of the current low level API. 

[...]

PR #443, opened 2023-02-04
Fix bug in TestSendDisembargo/SendQueuedResultToCaller

zenhack:
Sadly, I don't think this is #426, but I noticed it while hunting that down: we set Which to receiverHosted but then set the SenderHosted field. This patch fixes that.

I think we're getting lucky in the test such that it passes anyway because we only ever put one export on the wire so the ID is always zero anyway.

### DIFF
--- a/rpc/level1_test.go
+++ b/rpc/level1_test.go
@@ -255,8 +255,8 @@ func testSendDisembargo(t *testing.T, sendPrimeTo rpccp.Call_sendResultsTo_Which
 					Content: results.ToPtr(),
 					CapTable: []rpcCapDescriptor{
 						{
-							Which:        rpccp.CapDescriptor_Which_receiverHosted,
-							SenderHosted: importID,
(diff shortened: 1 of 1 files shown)


zenhack:
Have I mentioned I wish Go had sum types?

zenhack:
And indeed, this is not #426, since it triggered again. Re-running.

lthibault:
> Have I mentioned I wish Go had sum types?

It comes _so damn close_ with interfaces/type-switching, too.

lthibault (approved):

zenhack:
Quoting Louis Thibault (2023-02-04 14:04:30)

> It comes so damn close with interfaces/type-switching, too.

...We probably *could* make this safer by changing out the `Which` field
used by pogs to be an interface that you type-switch on, with nil just
being treated as the default. Would have to think about some of the
details, though I think this also gets easier if we implement #63, since
we then could move away from reflection, which I'd like to do anyway...

PR #446, opened 2023-02-09
BBR: avoid busy loop

zenhack:
...in the case where the connection is idle. Previously, if this branch
was taken then the select would always terminate immediately; we really
need it to block unless there's actually something to send.

This is probably a big part of the high cpu use we're seeing. Also: the requirement that the pipe be empty is probably why we are *not* seeing high CPU usage in tests.

---

Stacked on top of #445; review & merge that first.

### DIFF
--- a/flowcontrol/bbr/limiter.go
+++ b/flowcontrol/bbr/limiter.go
@@ -57,11 +57,6 @@ func (l *Limiter) computeBDP() float64 {
 }
 
 func (l *Limiter) run(ctx context.Context) {
-	// Dummy already-closed channel we can use when we want a channel
-	// that is ready, without needing to create it each time.
-	timeToSendNow := make(chan time.Time)
-	close(timeToSendNow)
-
 	for {
 
 		// These channels may or may not be nil, depending on what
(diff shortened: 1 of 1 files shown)


flowcontrol/bbr/limiter.go
```diff
@@ -84,9 +79,7 @@ func (l *Limiter) run(ctx context.Context) {
 			// might be in the far future, and there is no ack on its way
 			// to save us from our ignorance. Fortunately we always want
 			// to send if there's nothing on the wire, so just do it.
-			// Note that we don't use sendReqs here, since we don't want
-			// to skip over the logic that checks for app limited flows:
```
zenhack, line 88: I have no idea what my reasoning was with this comment, but it makes no sense as a justification, since either way the probability of choosing other branches of the select is non-zero.

I want to stare at this a bit more; I wonder if there's some bookkeeping that I knew we needed to be doing here, but apparently failed at actually ensuring we did? I'll have to have a look when we revisit fixing the estimation problems.

lthibault (approved):

PR #450, opened 2023-02-12
Streaming support helpers.

zenhack:
This patch begins work on ergonomic improvements to streaming workflows. The plan is to update the code generator to use SendStreamingCall for methods whose return type is stream, and have them similarly just return streamError instead of a result, release pair. This lets the caller just invoke the method in a loop and check the error each time through (or only at the end on a done() method or the like if they don't care about short-circuiting).

---

@lthibault this is untested, but I wanted to get your thoughts on the design before putting too much more work on this.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -125,7 +125,8 @@ type client struct {
 
 	mu          sync.Mutex // protects the struct
 	limiter     flowcontrol.FlowLimiter
-	streamError error       // Last error from streaming calls.
+	streamError error // Last error from streaming calls.
+	streamWg    *sync.WaitGroup
 	h           *clientHook // nil if resolved to nil or released
 	released    bool
 }
@@ -394,15 +395,26 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
 //    client will return the same error (without starting
 //    the method or calling PlaceArgs).
 func (c Client) SendStreamCall(ctx context.Context, s Send) error {
-	var streamError error
+	var (
+		streamError error
+		wg          *sync.WaitGroup
+	)
 	syncutil.With(&c.mu, func() {
 		streamError = c.streamError
+		if streamError == nil {
+			if c.streamWg == nil {
+				c.streamWg = &sync.WaitGroup{}
+			}
+			wg = c.streamWg
+			wg.Add(1)
+		}
 	})
 	if streamError != nil {
 		return streamError
 	}
 	ans, release := c.SendCall(ctx, s)
 	go func() {
+		defer wg.Done()
 		_, err := ans.Future().Ptr()
 		release()
 		if err != nil {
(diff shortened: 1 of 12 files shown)


capability.go
```diff
@@ -381,6 +387,33 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
 	return ans, rel
 }
 
+// SendStreamCall is like SendCall except that:
+//
+// 1. It does not return an answer for the eventual result.
+// 2. If the call returns an error, all future calls on this
+//    client will reutrn the same error. (without starting
```
lthibault, line 394: Typos: "retrun" & period before parentheses.

lthibault (changes requested):

lthibault:
The general direction looks good to me. At some point, it might be worthwhile to use a queue of some sort and avoid forking a goroutine on each call, but no need to worry about that yet.

zenhack:
Testing has revealed that as-written this doesn't work as I intended; there's a race where if your final call to done() or whatever goes out before the first error comes back in, you never see the error. I could try to rework things such that other calls actually check for errors on prior calls, but that seems messy; what if we just required the user to call a separate:

```
func (c Client) WaitStreaming(ctx context.Context) error
```
...which would wait for all outstanding streaming calls to settle (or for one to return an error), and then return the first error? So the pattern would be:

1. Make all your streaming calls. Check errors only if you care about short-circuiting.
2. Maybe call a Done() method or the like if that's part of your interface's contract.
3. Call WaitStreaming() to check if any of the streaming methods errored.

This seems like it would be much easier to implement than trying to instrument returned futures and the like, and it has the upside that it's now possible to distinguish between done() failing and a streaming call failing, though I'm not sure that's something that's likely to be needed.

Thoughts?

lthibault:
The `WaitStreaming` method and its associated pattern is analogous to how I've been doing streaming calls on the application side, so I'm onboard.  For what it's worth, I've found it convenient to write loops that look like the following:

```go
for stream.Open() {  // short-circuit?
    stream.DoTheThing(...)
}

if stream.Err() != nil {
    return stream.Err()
}
```

zenhack:
Great. Modulo further feedback, this is done.

capability.go
```diff
@@ -381,6 +388,65 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
...
+func (c Client) SendStreamCall(ctx context.Context, s Send) error {
+	var (
+		streamError error
+		wg          *sync.WaitGroup
+	)
+	syncutil.With(&c.mu, func() {
+		streamError = c.streamError
+		if streamError == nil {
+			if c.streamWg == nil {
+				c.streamWg = &sync.WaitGroup{}
```
lthibault, line 406: Is there a reason for this lazy initialization, as opposed to removing the pointer indirection on `streamWg`?

capability.go
```diff
@@ -381,6 +388,65 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
...
+		ret error
+		wg  *sync.WaitGroup
+	)
+	syncutil.With(&c.mu, func() {
+		wg = c.streamWg
+	})
+	if wg == nil {
+		return nil
+	}
+	wg.Wait()
```
lthibault, line 443: It seems like a lot of this logic could be abstracted away with `errgroup.Group`.  I'm fairly certain the zero-value `Group` will behave correctly, here.

Also, now that I think of it, shouldn't any outstanding calls to a streaming capability immediately fail if a previous call fails?  It seems like `errgroup.WithContext` could help with this.

capability.go
```diff
@@ -324,8 +327,8 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
 		return ErrorAnswer(s.Method, errors.New("call on null client")), func() {}
 	}
 
-	if c.streamError != nil {
-		return ErrorAnswer(s.Method, exc.WrapError("stream error", c.streamError)), func() {}
+	if c.stream.err != nil {
+		return ErrorAnswer(s.Method, exc.WrapError("stream error", c.stream.err)), func() {}
```
lthibault, line 331: Do we need to hold the lock to access `c.stream.err`?

lthibault (approved):

Issue #348, opened 2022-11-24
panic (close of closed channel) in TestCallReceiverRpc


zenhack:
Per #346; haven't dug into this at all yet, but I assume this has been there and is just rare enough to have not cropped up before. It passed ~250 times in that run before failing...

https://github.com/capnproto/go-capnproto2/actions/runs/3537277747/jobs/5937100516#step:7:488

```
--- FAIL: TestCallReceiverAnswerRpc (0.00s)
panic: close of closed channel [recovered]
	panic: close of closed channel

goroutine 39 [running]:
testing.tRunner.func1.2({0x829bd80, 0x8334130})
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1396 +0x2ab
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1399 +0x41f
panic({0x829bd80, 0x8334130})
	/opt/hostedtoolcache/go/1.19.3/x64/src/runtime/panic.go:884 +0x1c3
capnproto.org/go/capnp/v3.Client.Release({0x8c604b0})
	/home/runner/work/go-capnproto2/go-capnproto2/capability.go:584 +0xe6
capnproto.org/go/capnp/v3.(*Message).Reset(0x8c64540, {0x0, 0x0})
	/home/runner/work/go-capnproto2/go-capnproto2/message.go:134 +0xc6
capnproto.org/go/capnp/v3/rpc/transport.(*transport).RecvMessage.func1()
	/home/runner/work/go-capnproto2/go-capnproto2/rpc/transport/transport.go:183 +0x30
capnproto.org/go/capnp/v3/rpc.(*question).PipelineSend.func3()
	/home/runner/work/go-capnproto2/go-capnproto2/rpc/question.go:182 +0x4e
capnproto.org/go/capnp/v3/rpc.TestCallReceiverAnswerRpc(0x8c97770)
	/home/runner/work/go-capnproto2/go-capnproto2/rpc/receiveranswer_test.go:137 +0x583
testing.tRunner(0x8c97770, 0x82e0b2c)
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1446 +0x113
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.3/x64/src/testing/testing.go:1493 +0x374
exit status 2
FAIL	capnproto.org/go/capnp/v3/rpc	0.115s
[...]

PR #452, opened 2023-02-15
WIP: sketch out interfaces for working with vat networks

zenhack:
Similar to the VatNetwork interfaces described at the bottom of rpc.capnp; this patch introduces a sketch of what these interfaces might look like in Go, plus the beginning an in-memory implementation to use in tests. Needs testing & docs, and of course isn't actually used by anything.


zenhack:
Note this is using the term "Peer" instead of "Vat" in the API, because the latter is really a unit of concurrency, not a node in the network. See https://github.com/ocapn/ocapn/issues/18

lthibault:
I like this.  I've stared at that spec a few times, and have often wanted an implementation to play around with.

One thing I'm still not clear about, though, is what bearing this has (if any) on how applications interact with the capnp library.  Are application developers going to need to provide their own implementation of these interfaces in order to support 3PH?  If so, I would appreciate a quick sketch of how this interacts with the existing `rpc.Conn` API.  I'm sure it's simple enough, but it just hasn't clicked yet.

zenhack:
Quoting Louis Thibault (2023-02-15 14:11:27)

> Are application developers going to need to provide their own
> implementation of these interfaces in order to support 3PH?

We'll probably want to define and implement at least one that people can
use out of the box. See my comments about this on the 3PH issue.

> I would appreciate a quick sketch of how this interacts with the
> existing rpc.Conn API. I'm sure it's simple enough, but it just hasn't
> clicked yet.

Basically, instead of supplying a Transport to NewConn, you'd
construct a Network spawn a goroutine calling Accept() in a loop,
and optionally call Dial to connect to some other peer and get its
bootstrap interface.

This bit should be documented (hence this is WIP) but: the idea
with the Network and PeerId fields on on rpc.Options is that
the logic inside Conn would use these when dealing with third
party caps and Provide/Accept message types. Dial() and Accept()
on Network would fill these fields in themselves; app code wouldn't
generally deal with them directly.

zenhack:
(update, went ahead and documented those fields).

rpc/rpc.go
```diff
@@ -192,8 +192,17 @@ type Options struct {
 	// timeout is used.
 	AbortTimeout time.Duration
 
+	// RemotePeerId is the PeerId of the remote side of the connection. Can
+	// be left as the zero value for point to point connections. For >= 3
+	// party use, this should be filled in by the Network on Accept or Dial.
+	// Application code should not set this.
+	RemotePeerId PeerId
```
lthibault, line 199: nit:  I think we should format this as `RemotePeerID`/`PeerID`

lthibault (changes requested):

rpc/network.go
```diff
@@ -0,0 +1,29 @@
+package rpc
+
+import (
+	"context"
+
+	capnp "capnproto.org/go/capnp/v3"
+)
+
+type PeerId struct {
```
lthibault, line 9: Is it reasonable to remove this and make the interfaces accept a type parameter instead?  It would be quite elegant to be able to reuse libp2p's `peer.ID` type, and I think it would also make sense for the type system to distinguish between, say, `Network[peer.ID]` and `Network[uuid.UUID]`.

zenhack:
Quoting Louis Thibault (2023-02-15 15:25:03)

>    Is it reasonable to remove this and make the interfaces accept a
>    type parameter instead? It would be quite elegant to be able to
>    reuse libp2p's peer.ID type, and I think it would also make sense
>    for the type system to distinguish between, say, Network[peer.ID]
>    and Network[uuid.UUID].

I considered that, but we're going to want to store these inside a Conn,
so we're going to end up needing to hide the type anyway. I suppose we
could make the PeerId wrapper private though? But then it's still going
to propagate into Options, so you'd have to specify a type for that even
when doing point-to-point connections...

zenhack:
Added a bunch of docs.

lthibault:
>I considered that, but we're going to want to store these inside a Conn,
so we're going to end up needing to hide the type anyway. I suppose we
could make the PeerId wrapper private though? But then it's still going
to propagate into Options, so you'd have to specify a type for that even
when doing point-to-point connections...

Ah ok, that makes sense.

Partly as an intellectual exercise, and partly in case there's something there:  the options type is consumed internally by `rpc.Conn`, if I'm not mistaken.  What I mean is that none of `Conn`s methods will return a parametrized type (right?).

If that is indeed the case, it seems like we could get away with defining `Conn` to be an interface that is satsified by the parametrized (and possibly private) conn-struct.

I have a hunch that it will be helpful to enforce separation of vat-networks based on type.  Maybe I'm overthinking this, though.

zenhack:
#348, re-running.

zenhack:
That feels overly complex. I kindof suspect applications that use more than one Network will be relatively rare, and also I see various problems with things like importClient which hold a pointer to the connection; you'd have to again propagate the type parameter... it just doesn't seem worth it to head off mixing up peer IDs.

I'm not 100% happy with this; I think there are tweaks I'll want to make to the interface, but it's at the point where it's getting hard to predict exactly what they will be before diving in and actually trying to implement the logic this is supposed to support -- what do you say to merging this as-is with the understanding that it will probably be revised a few times?

rpc/network.go
```diff
@@ -0,0 +1,95 @@
...
+	SendToRecipient ThirdPartyCapID
+	SendToProvider  RecipientID
+}
+
+// A Network is a reference to a multi-party (generally >= 3) network
+// of Cap'n Proto peers. Use this instead of NewConn when establishing
+// connections outside a point-to-point setting.
+type Network interface {
+	// Return the identifier for caller on this network.
+	MyID() PeerID
```
lthibault, line 67: `LocalID`?  Or how about just `ID`?

lthibault (changes requested):
Yeah, that sounds good. 

Code looks good overall, modulo a small naming nit.

rpc/network.go
zenhack: Done.

lthibault (approved):

PR #455, opened 2023-02-18
WIP: groundwork for request-based API

zenhack:
...as discussed in #64. To avoid an extra copy we will have to rework ClientHook & possibly other low level APIs, but this will let us play with code generation and user-facing APIs.

### DIFF
--- a/request.go
+++ b/request.go
@@ -1,0 +1,81 @@
+package capnp
+
+import (
+	"context"
+)
+
+type Request struct {
+	method          Method
+	args            Struct
+	client          Client
+	releaseResponse ReleaseFunc
+	future          *Future
+}
+
+func NewRequest(client Client, method Method, argsSize ObjectSize) (*Request, error) {
+	_, seg, err := NewMessage(MultiSegment(nil))
+	if err != nil {
+		return nil, err
+	}
+	args, err := NewStruct(seg, argsSize)
+	if err != nil {
+		return nil, err
+	}
+	return &Request{
+		method: method,
+		args:   args,
+		client: client,
+	}, nil
+}
+
+func (r *Request) Args() Struct {
+	return r.args
+}
+
+func (r *Request) getSend() Send {
+	return Send{
+		Method: r.method,
+		PlaceArgs: func(args Struct) error {
+			err := args.CopyFrom(r.args)
+			r.releaseArgs()
+			return err
+		},
+		ArgsSize: r.args.Size(),
+	}
+}
+
+func (r *Request) Send(ctx context.Context) *Future {
+	ans, rel := r.client.SendCall(ctx, r.getSend())
+	r.releaseResponse = rel
+	r.future = ans.Future()
+	return r.future
+}
+
+func (r *Request) SendStream(ctx context.Context) error {
+	return r.client.SendStreamCall(ctx, r.getSend())
+}
+
+func (r *Request) Future() *Future {
+	return r.future
+}
+
+func (r *Request) Release() {
+	r.releaseArgs()
+	rel := r.releaseResponse
+	if rel != nil {
+		r.releaseResponse = nil
+		r.future = nil
+		rel()
+	}
+}
+
+func (r *Request) releaseArgs() {
+	if r.args.IsValid() {
+		return
+	}
+	msg := r.args.Message()
+	r.args = Struct{}
+	arena := msg.Arena
+	msg.Reset(nil)
+	arena.Release()
+}
--- a/request.go
+++ b/request.go
@@ -2,6 +2,7 @@ package capnp
 
 import (
 	"context"
+	"errors"
 )
 
 // A Request is a method call to be sent. Create one with NewReqeust, and send it with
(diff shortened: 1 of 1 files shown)


request.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, err
+	}
+	return &Request{
+		method: method,
+		args:   args,
+		client: client,
+	}, nil
+}
+
+func (r *Request) Args() Struct {
```
lthibault, line 31: Any reason not to just export the `args` field?  Are we planning on passing this through an interface?

request.go
```diff
@@ -0,0 +1,81 @@
...
+	r.releaseResponse = rel
+	r.future = ans.Future()
+	return r.future
+}
+
+func (r *Request) SendStream(ctx context.Context) error {
+	return r.client.SendStreamCall(ctx, r.getSend())
+}
+
+func (r *Request) Future() *Future {
```
lthibault, line 58: Same question here

request.go
zenhack: The main reason is we'll end up with typed wrappers for this, so we can't just change the type of the field in the wrapper, but we *can* override the method with one that casts. I considered an alternative where the type was generic in the parameter and return type:

```go
type Request[Param ~StructKind, Result_Future ~FutureKind] struct {
    Args Param
    Future Results_Future
    // ...
}
```

But esp. when thinking through the details it felt much more complicated. The asymmetry between the parameter (struct) and result (future) also felt really ugly and unintuitive, and we don't otherwise need to have a `FutureKind` defined. I also think it would make for more verbose call sites.

Another minor arguments in favor, this is more consistent with the structs that get passed to method implementations (which also have an `.Args()`) method.

lthibault: Makes sense 👍.

request.go
```diff
@@ -0,0 +1,81 @@
...
+type Request struct {
+	method          Method
+	args            Struct
+	client          Client
+	releaseResponse ReleaseFunc
+	future          *Future
+}
+
+func NewRequest(client Client, method Method, argsSize ObjectSize) (*Request, error) {
+	_, seg, err := NewMessage(MultiSegment(nil))
```
lthibault, line 16: `NewMultiSegmentMessage(nil)`?  I know we're able to return an error here, but since it should never fail, may as well?

(Alternatively, maybe we want to take an `Arena` as an argument?  Or is that a footgun?)

lthibault (commented):
I forgot I reviewed this a few days ago, and read through it again.  But hey, at least I saw a thing I overlooked last time!

request.go
zenhack: Good catch. Yeah, I'd forgotten that function exists until the other day. Done.

lthibault:
If this is "done" for you, then it LGTM.  Happy to re-review later if you want to add more, of course.

zenhack:
It could probably at least use some docs before we merge this piece.

lthibault (approved):
Feel free to merge when ready (unless you want me to review the docs).

lthibault (approved):
Just a thought:  what happens if I call `Send()` multiple times on the same request?  Do we want to make it idempotent?

lthibault:
Then again, it might be useful in cases where you want to repeatedly make the same call?  I wouldn't want to support this if it makes implementation overly complex, of course.

I'm also fine with the simple approach of documenting behavior, e.g. "Callers MUST NOT call Send more than once."

zenhack:
That gets weird since we store the releasefunc and future -- would we have a slice of them? seems not worth it.

Checking for a double-call is easy, so we may as well; I'll do that.

lthibault:
SGTM.  👍

zenhack:
Done.

lthibault (approved):

request.go
```diff
@@ -0,0 +1,102 @@
...
+			return err
+		},
+		ArgsSize: r.args.Size(),
+	}
+}
+
+// Send sends the request, returning a future for its results.
+func (r *Request) Send(ctx context.Context) *Future {
+	if r.future != nil {
+		return ErrorAnswer(r.method, errors.New("Sent the same request twice.")).Future()
```
lthibault, line 54: We shouldn't capitalize or punctuate error messages.

lthibault (changes requested):

request.go
```diff
@@ -0,0 +1,102 @@
...
+	ans, rel := r.client.SendCall(ctx, r.getSend())
+	r.releaseResponse = rel
+	r.future = ans.Future()
+	return r.future
+}
+
+// SendStream is to send as Client.SendStreamCall is to Client.SendCall
+func (r *Request) SendStream(ctx context.Context) error {
+	if r.future != nil {
+		return errors.New("Sent the same request twice.")
```
lthibault, line 66: Ditto

lthibault (changes requested):

request.go
zenhack: Done.

lthibault (approved):

PR #456, opened 2023-02-19
Avoid use of fmt in the main package.

zenhack:
This adds a String() method to ClientHook so we can avoid using fmt for that, then removes the last use of fmt in the main package, per #364.

Before closing that issue, I think I'd still like to get it out of the flowcontrol package hierarchy at least, and maybe elsewhere if practical.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -540,6 +540,13 @@ func (pc PipelineClient) Brand() Brand {
 func (pc PipelineClient) Shutdown() {
 }
 
+func (pc PipelineClient) String() string {
+	return "PipelineClient{transform: " +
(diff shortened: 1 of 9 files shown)


zenhack:
I was mistaken, there is at least one more use of fmt in the main package (in list.go)

zenhack:
Test failures were because I forgot to add a String() for dummyHook; fixed.

zenhack:
Added another commit that removes the dep from flowcontrol as well.

zenhack:
...apparently I'm bad at remembering to test before pushing today. Fixed.

lthibault (approved):

PR #458, opened 2023-02-20
WIP: avoid deadlocks wrt. embargo

zenhack:
This patch:

- Moves the answerQueue (and structReturner) types from the server package into the main package, and exports them.
- Reworks the implementation of embargo in the rpc package to use answerQueue; this makes the queue unbounded instead of having Send & Recv just block, which should head off some potential deadlocks which right now I think require multiple connections to trigger.

Marking this WIP because there are a few bits that are still a little dodgy, both stylistically and in terms of correctness. But the tests pass.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -270,12 +270,12 @@ func ErrorAnswer(m Method, e error) *Answer {
 	return &p.ans
 }
 
-// ImmediateAnswer returns an Answer that accesses s.
-func ImmediateAnswer(m Method, s Struct) *Answer {
+// ImmediateAnswer returns an Answer that accesses ptr.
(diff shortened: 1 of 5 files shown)


lthibault (approved):

zenhack:
Ok, this is ready now.

zenhack:
I have no idea what to make of this: https://github.com/capnproto/go-capnproto2/actions/runs/4259670598/jobs/7412098591#step:5:636

It looks like a compile error? I'm going to re-run, but that's bizzare either way. It *definitely* builds locally.

lthibault (approved):

PR #463, opened 2023-02-25
Remove roSingleSegmentArena

lthibault:
Removes `roSingleSegmentArena`, using pooled `SingleSegmentArena` instances instead.  Performance should be comparable.  Simplifies segment reuse logic, and performs minor rearranging of `Message.Decode` to improve readability.

No functional changes.

Stacks on top of #462.  Merge that first.


message.go
```diff
@@ -699,9 +676,9 @@ type Decoder struct {
 	bufferPool *bufferpool.Pool
 
 	reuse bool
-	buf   []byte
+	// buf   []byte
```
zenhack: Should delete this, rather than commenting it out.

message.go
```diff
@@ -771,42 +747,30 @@ func (d *Decoder) Decode() (*Message, error) {
...
-		if err != nil {
-			return nil, exc.WrapError("decode", err)
-		}
-		return &Message{
-			Arena:          arena,
-			originalBuffer: buf,
-		}, nil
+	if d.reuse {
+		d.arena.Release()
+		d.arena = singleSegmentPool.Get().(Arena)
```
zenhack, line 752: Any reason not to just call `SingleSegment(nil)`?

message.go
```diff
@@ -771,42 +747,30 @@ func (d *Decoder) Decode() (*Message, error) {
...
+		d.msg.Reset(d.arena)
+		return &d.msg, nil
```
zenhack, line 754: Wait, do we not read the message in this case?

zenhack (changes requested):

message.go
lthibault: Fixed.

PR #465, opened 2023-02-25
Allocate SingleSegmentArena from bufferpool.

lthibault:
Allocates and resizes single-segment arena using default buffer pool.  Note:  buffer pool also takes care of zeroing out memory.


zenhack:
Hm, #348 again. I wonder if this mucks with the timing enough to make that more likely?

zenhack (approved):

PR #466, opened 2023-02-25
Improvements to buffer pool

lthibault:
- [x] Removes unused buffer pool buckets
- [x] Improves buffer reuse by consuming buffers even if capacity is not aligned to bucket boundary
- [x] Aggressively reuses small-sized buffers by setting min page size to 1kb
- [x] Improves buffer reuse by decreasing total number of buckets
- [x] Small readability improvements (use `range` instead of classic `for` loop)
- [x] Increase max page size to 1Mb

### DIFF
--- a/exp/bufferpool/pool.go
+++ b/exp/bufferpool/pool.go
@@ -1,52 +1,126 @@
 // Package bufferpool supports object pooling for byte buffers.
 package bufferpool
 
-import "sync"
+import (
+	"sync"
+)
 
 const (
-	nBuckets = 20
+	defaultMinSize     = 1024
+	defaultBucketCount = 20
 )
 
 // A default global pool.
 var Default Pool
 
-// A pool of buffers, in variable sizes.
+// Pool maintains a list of BucketCount buckets that contain buffers
+// of exponentially-increasing capacity, 1 << 0 to 1 << BucketCount.
+//
+// The MinAlloc field specifies the minimum capacity of new buffers
+// allocated by Pool, which improves reuse of small buffers. For the
+// avoidance of doubt:  calls to Get() with size < MinAlloc return a
+// buffer of len(buf) = size and cap(buf) >= MinAlloc. MinAlloc MUST
+// NOT exceed 1 << BucketCount, or method calls to Pool will panic.
+//
+// The zero-value Pool is ready to use, defaulting to BucketCount=20
+// and MinAlloc=1024 (max size = ~1MiB).  Most applications will not
+// benefit from tuning these parameters.
+//
+// As a general rule, increasing MinAlloc reduces GC latency at the
+// expense of increased memory usage.  Increasing BucketCount can
+// reduce GC latency in applications that frequently allocate large
+// buffers.
 type Pool struct {
-	buckets [nBuckets]sync.Pool
+	once                  sync.Once
+	MinAlloc, BucketCount int
+	buckets               bucketSlice
 }
 
-// Get a buffer of the given length. Its capacity may be larger than the
-// requested size.
+// Get a buffer of len(buf) == size and cap >= size.
 func (p *Pool) Get(size int) []byte {
-	for i := 0; i < nBuckets; i++ {
-		capacity := 1 << i
-		if capacity >= size {
-			var ret []byte
-			item := p.buckets[i].Get()
-			if item == nil {
-				ret = make([]byte, capacity)
-			} else {
-				ret = item.([]byte)
-			}
-			ret = ret[:size]
-			return ret
-		}
+	p.init()
+
+	if buf := p.buckets.Get(size); buf != nil {
+		return buf[:size]
 	}
+
 	return make([]byte, size)
 }
 
-// Return a buffer to the pool. Zeros the slice (but not the full backing array)
-// before making it available for future use.
+// Put returns the buffer to the pool.  The first len(buf) bytes
+// of the buffer are zeroed.
 func (p *Pool) Put(buf []byte) {
-	for i := 0; i < len(buf); i++ {
+	p.init()
+
+	for i := range buf {
 		buf[i] = 0
 	}
 
-	capacity := cap(buf)
-	for i := 0; i < nBuckets; i++ {
-		if (1 << i) == capacity {
-			p.buckets[i].Put(buf[:capacity])
-			return
+	p.buckets.Put(buf[:cap(buf)])
+}
+
+func (p *Pool) init() {
+	p.once.Do(func() {
+		if p.MinAlloc <= 0 {
+			p.MinAlloc = defaultMinSize
 		}
+
+		if p.BucketCount <= 0 {
+			p.BucketCount = defaultBucketCount
+		}
+
+		if p.MinAlloc > (1 << p.BucketCount) {
+			panic("MinAlloc greater than largest bucket")
+		}
+
+		// Get the index of the bucket responsible for MinAlloc.
+		var idx int
+		for idx = range p.buckets {
+			if 1<<idx >= p.MinAlloc {
+				break
+			}
+		}
+
+		p.buckets = make(bucketSlice, p.BucketCount)
+		for i := range p.buckets {
+			if i < idx {
+				// Set the 'New' function for all "small" buckets to
+				// n.buckets[idx].Get, so as to allow reuse of buffers
+				// smaller than MinAlloc that are passed to Put, while
+				// still maximizing reuse of buffers allocated by Get.
+				// Note that we cannot simply use n.buckets[idx].New,
+				// as this would side-step pooling.
+				p.buckets[i].New = p.buckets[idx].Get
+			} else {
+				p.buckets[i].New = newAllocFunc(i)
+			}
+		}
+	})
+}
+
+type bucketSlice []sync.Pool
+
+func (bs bucketSlice) Get(size int) []byte {
+	for i := range bs {
+		if 1<<i >= size {
+			return bs[i].Get().([]byte)
+		}
+	}
+
+	return nil
+}
+
+func (bs bucketSlice) Put(buf []byte) {
+	for i := range bs {
+		if cap(buf) >= 1<<i && cap(buf) < 1<<(i+1) {
+			bs[i].Put(buf)
+			break
+		}
+	}
+}
+
+func newAllocFunc(i int) func() any {
+	return func() any {
+		return make([]byte, 1<<i)
 	}
 }
--- a/exp/bufferpool/pool_test.go
+++ b/exp/bufferpool/pool_test.go
@@ -1,0 +1,75 @@
+package bufferpool_test
+
+import (
+	"testing"
+
+	"capnproto.org/go/capnp/v3/exp/bufferpool"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestInvariants(t *testing.T) {
+	t.Parallel()
+
+	pool := bufferpool.Pool{
+		BucketCount: 1,
+		MinAlloc:    1024,
+	}
+
+	assert.Panics(t, func() {
+		pool.Get(32)
+	}, "should panic when BucketCount cannot satisfy MinAlloc.")
+}
+
+func TestGet(t *testing.T) {
+	t.Parallel()
+	t.Helper()
+
+	pool := bufferpool.Pool{
+		BucketCount: 8,
+		MinAlloc:    64,
+	}
+
+	t.Run("Size<MinAlloc", func(t *testing.T) {
+		assert.Len(t, pool.Get(32), 32, "should return buffer with len=32")
+		assert.Equal(t, 32, cap(pool.Get(32)), "should return buffer with cap=MinAlloc")
+	})
+
+	t.Run("Size=MinAlloc", func(t *testing.T) {
+		assert.Len(t, pool.Get(pool.MinAlloc), pool.MinAlloc, "should return buffer with len=MinAlloc")
+		assert.Equal(t, pool.MinAlloc, cap(pool.Get(pool.MinAlloc)), "should return buffer with cap=MinAlloc")
+	})
+
+	t.Run("Size>MinAlloc", func(t *testing.T) {
+		assert.Len(t, pool.Get(33), 33, "should return buffer with len=33")
+		assert.Equal(t, 128, cap(pool.Get(128)), "should return buffer with cap=128")
+	})
+
+	t.Run("Size>MaxSize", func(t *testing.T) {
+		assert.Len(t, pool.Get(512), 512, "should return buffer with len=512")
+		assert.GreaterOrEqual(t, 512, cap(pool.Get(512)), "should return buffer with cap>=512")
+	})
+}
+
+func TestPut(t *testing.T) {
+	t.Parallel()
+
+	var pool bufferpool.Pool
+
+	buf := make([]byte, 1024)
+	for i := range buf {
+		buf[i] = byte(i)
+	}
+	buf = buf[:8]
+
+	pool.Put(buf)
+	buf = pool.Get(8)
+
+	assert.Equal(t, make([]byte, 8), buf, "should zero first 8 bytes")
+}
+
+func BenchmarkPool(b *testing.B) {
+	for i := 0; i < b.N; i++ {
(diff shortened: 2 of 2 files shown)


message.go
```diff
@@ -465,8 +462,9 @@ func (ssa *SingleSegmentArena) Allocate(sz Size, segs map[SegmentID]*Segment) (S
 	if err != nil {
 		return 0, nil, err
 	}
-	buf := make([]byte, len(data), cap(data)+inc)
-	copy(buf, data)
+	buf := bufferpool.Default.Get(cap(data) + inc)
+	buf = buf[:copy(buf, data)]
```
zenhack, line 466: Shoving all this on one line makes it hard to follow imo; maybe:

```
copied := copy(buf, data)
buf = buf[:copied]
```

exp/bufferpool/pool.go
```diff
@@ -18,33 +19,29 @@ type Pool struct {
...
-	for i := 0; i < len(buf); i++ {
+	for i := range buf {
 		buf[i] = 0
 	}
 
 	capacity := cap(buf)
-	for i := 0; i < nBuckets; i++ {
-		if (1 << i) == capacity {
+	for i := range p.buckets {
+		if (1 << i) >= capacity {
```
zenhack, line 44: - Changing this to `>=` is going to result in undersized buffers in the buckets, which is going to break things if someone asks for a buffer bigger than what's returned but <= the pool size, unless I'm missing something.
- Do we need to change `1 << i` to `minSize << i` here as well?

...we should probably add some tests for this package, so we catch this kind of thing.

message.go
```diff
@@ -485,27 +483,8 @@ func (ssa SingleSegmentArena) String() string {
...
-func singleSegment(b []byte) *SingleSegmentArena {
-	return (*SingleSegmentArena)(&b)
-}
-
-var singleSegmentPool = sync.Pool{
-	New: func() any {
-		return singleSegment(nil)
-	},
+	bufferpool.Default.Put(*ssa)
+	(*ssa) = nil
```
zenhack, line 487: Nit: don't need the parens on the lvalue

exp/bufferpool/pool.go
```diff
@@ -4,7 +4,8 @@ package bufferpool
 import "sync"
 
 const (
-	nBuckets = 20
+	minSize  = 1024
+	nBuckets = 11 // max 1Mb
```
zenhack, line 8: Maybe spell out the math behind this a bit more, or (better, if you can figure out how) somehow define it as an expression.

Also, nit: the unit you're looking for is MiB.

Also, I wish Go had static assertions.

zenhack (changes requested):
A couple things in-line

exp/bufferpool/pool.go
lthibault: >Also, nit: the unit you're looking for is MiB.

Thanks - for some reason, I can never remember this 😆

exp/bufferpool/pool.go
lthibault: Ugh, this is a late-night mistake.  The intent was to allow buffers larger than the bucket's size, but smaller than the _next_ bucket's size.  Obviously this code doesn't do that, though, so I'll fix it.

message.go
lthibault: This shouldn't have been pushed to this PR.  I've corrected it in #465.

message.go
lthibault: Ditto. Fixed in #465.

exp/bufferpool/pool.go
```diff
@@ -18,33 +19,32 @@ type Pool struct {
...
+	for i := range buf {
 		buf[i] = 0
 	}
 
 	capacity := cap(buf)
-	for i := 0; i < nBuckets; i++ {
-		if (1 << i) == capacity {
+	for i := range p.buckets {
+		bucket := (minSize << i)
+		next := (minSize<<i + 1)
```
zenhack, line 45: Maybe `minSize << (i + 1)`? This might be right, but I can't remember the precedence rules well enough without looking it up, and it would be clearer I think.

Also the parens on the first one are entirely superfluous.

zenhack (changes requested):
One more minor style nit

lthibault:
I ended up re-writing a significant portion of this to allow users to provide their own pool implementations.  I think this can have a very large impact in niche cases (e.g. embedded systems), where you might want to tune the pooling behavior to save memory.  I can also imagine applications that allocate a large volume of very big segments.

I've also added tests around the sensitive parts of the default pool implementation.

capability.go
```diff
@@ -397,10 +397,10 @@ func (c Client) SendCall(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
 
 // SendStreamCall is like SendCall except that:
 //
-// 1. It does not return an answer for the eventual result.
-// 2. If the call returns an error, all future calls on this
-//    client will return the same error (without starting
-//    the method or calling PlaceArgs).
+//  1. It does not return an answer for the eventual result.
```
lthibault, line 400: Not sure what happened here.  gofmt, maybe?

lthibault:
Weird, I'm getting [build failures](https://github.com/capnproto/go-capnp/actions/runs/4271716582/jobs/7436377807).  Re-running 🤨

lthibault:
Ok fixed. Ready for review.

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
-			item := p.buckets[i].Get()
-			if item == nil {
-				ret = make([]byte, capacity)
-			} else {
-				ret = item.([]byte)
-			}
-			ret = ret[:size]
-			return ret
-		}
+var pool Pool = &BasicPool{}
```
zenhack, line 11: Is there a reason for this to no longer be public?

I might name it `defaultPool`; `pool` is a bit too generic. Or keep it as `Default` if it's going to be exported.

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
+func Set(p Pool) {
+	if p == nil {
+		panic("Pool(nil)")
 	}
-	return make([]byte, size)
+
+	pool = p
+}
+
+// Get a buffer of with the specified size from the global pool.  The
```
zenhack, line 31: Grammar.

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
+// resulting buffer's capacity MAY exceed size. The buffer SHOULD NOT
+// be resized beyond its underlying capacity, or it will leak to the
+// garbage collector.
```
zenhack, line 34: I might just replace all this with something to the effect of `calls .Get() on the global Pool`, maybe moving some of the details here to that doc comment. Same thing applies to Put()

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
+// Put returns the buffer to the global pool.  The slice will be zeroed
+// before returning it to the pool.   Note that only the first len(buf)
+// bytes are zeroed, and not the full backing array.
+//
+// TODO:  should we zero out the entire array?  Users may have resized
+// the slice, so it may contain sensitive data.
+func Put(buf []byte) {
+	for i := range buf {
 		buf[i] = 0
 	}
```
zenhack, line 48: Should we be doing this here, or in the implementation of Pool.Put?

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
-	capacity := cap(buf)
-	for i := 0; i < nBuckets; i++ {
-		if (1 << i) == capacity {
-			p.buckets[i].Put(buf[:capacity])
+	pool.Put(buf[:cap(buf)])
+}
+
+// Pool is a free list of []byte buffers of heterogeneous sizes.
+// Implementations MUST be thread safe and MUST support efficient
+// paging of buffers with arbitrary capacity.   Most applications
```
zenhack, line 57: I'm not exactly sure what paging means in this context.

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,161 @@ package bufferpool
...
+// and Min=1024 (1KiB - ~1MiB buffer sizes).  Most applications
+// will not benefit from tuning these parameters.
+//
+// As a general rule, increasing Min reduces GC latency at the
+// expense of increased memory usage.  Increasing N can reduce
+// GC latency in applications that frequently allocate buffers
+// of size >=1MiB.
+type BasicPool struct {
+	once    sync.Once
+	Min, N  int
```
zenhack, line 86: Maybe `MinSize` and `BucketCount`? `N` in particular as a field name feels not descriptive enough.

zenhack (changes requested):
Some comments in-line

exp/bufferpool/pool.go
lthibault: Two reasons:

1. avoid it being set to `nil`
2. force calls to go through the top-level `Get` and `Put` methods, since they handle things like zeroing memory, which keeps the interface implementation simple.

lthibault: >I might name it defaultPool; pool is a bit too generic. 

Done.

exp/bufferpool/pool.go
lthibault: I personally like it here, as it keeps the implementation of `Pool` simpler.  It also prevents user-supplied implementations from forgetting to do this.

exp/bufferpool/pool.go
lthibault: Done.

exp/bufferpool/pool.go
lthibault: Fixed (dunno what I was thinking).

exp/bufferpool/pool.go
lthibault: @zenhack what do you think of the TODO?  Any reason why we're not zeroing out the entire array?  Is it out of concern for performance?

exp/bufferpool/pool.go
lthibault: Done

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,151 @@ package bufferpool
...
+	defaultPool = p
+}
+
+// Get calls .Get() on the global pool, and resizes the resulting buf
+// to size.
+func Get(size int) []byte {
+	return defaultPool.Get(size)[:size]
+}
+
+// Put calls .Put(buff) on the global pool after zeroing the supplied
```
zenhack, line 37: I really think we should move the zeroing logic back into Pool.Put, *especially* now that the interface is gone. Same thing with the capacity check.

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,151 @@ package bufferpool
...
+	}
+}
+
+// Pool maintains a list of BucketCount buckets containing buffers
+// of exponentially-increasing size, starting from MinSize. Buffers
+// whose capacity exceeds MinSize << BucketCount are ignored and left
+// to the garbage-collector.
+//
+// The zero-value Pool is ready to use, defaulting to BucketCount=10
+// and MinSize=1024 (1KiB - ~1MiB buffer sizes).  Most applications
```
zenhack, line 59: Nit: the tilde is unnecessary; Mi is a power of two prefix, so this is actually exact (the base 10 prefixes don't have the i).

exp/bufferpool/pool.go
```diff
@@ -4,49 +4,151 @@ package bufferpool
...
-	buckets [nBuckets]sync.Pool
+	once                 sync.Once
+	MinSize, BucketCount int
+	buckets              []bucket
 }
 
-// Get a buffer of the given length. Its capacity may be larger than the
-// requested size.
+// Get a buffer of len >= size.  The caller will resize the
+// buffer to size, so implementations need not bother.  The
```
zenhack, line 73: The "implementations..." language should probably change now that this is not an interface.

zenhack (changes requested):

exp/bufferpool/pool.go
lthibault: Unless my arithmetic is mistaken (totally possible), the slices contained in the largest bucket will have a capacity of at most 1048575 Bytes (.99 MiB).  See `TryPut`:

```
func (b bucket) TryPut(buf []byte) (ok bool) {
	if ok = b.minCap <= cap(buf) && b.capLimit > cap(buf); ok {
		b.pool.Put(buf)
	}

	return
}
```

exp/bufferpool/pool.go
lthibault: If we're not splitting the zeroing vs pooling logic, then I see no reason to make the default pool a private variable.  I'll revert back to what we had before and remove the top-level `Get` and `Put` functions.

lthibault: Done.

exp/bufferpool/pool.go
zenhack: Bucket i should have capacity `1024 << i` = `1024 * 2 ^ i` = `2 ^ 10 * 2 * i` = `2 ^ (10 + i)` so with 10 buckets the largest should be `2 ^ 20` = 1MiB. What am I missing?

zenhack: Actually, with 10 buckets the last bucket is `i = 9`, so the limit is actually half a MiB...

exp/bufferpool/pool.go
```diff
@@ -1,34 +1,38 @@
...
+// avoidance of doubt:  calls to Get() with size < MinAlloc return a
+// buffer of len(buf) = size and cap(buf) >= MinAlloc. MinAlloc MUST
+// NOT exceed 1 << BucketCount, or method calls to Pool will panic.
 //
-// The zero-value Pool is ready to use, defaulting to BucketCount=10
-// and MinSize=1024 (1KiB - ~1MiB buffer sizes).  Most applications
-// will not benefit from tuning these parameters.
+// The zero-value Pool is ready to use, defaulting to BucketCount=20
+// and MinAlloc=1024 (max size = ~1MiB).  Most applications will not
+// benefit from tuning these parameters.
```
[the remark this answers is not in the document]
lthibault: Did you see my message on Matrix?  I ended up changing the implementation a bit.
(Apologies for the noise.  The scope of this PR ended up shifting around more than I'd like.  Maybe I should close it and open a fresh one?)

exp/bufferpool/pool.go
```diff
@@ -1,52 +1,135 @@
...
+		for idx = range p.buckets {
+			if 1<<idx >= p.MinAlloc {
+				break
+			}
+		}
+
+		// Set the 'New' function for all buckets lower than 'MinSize'
+		// to the MinSize bucket's 'New' function.  This ensures that
+		// we always allocate a capacity of *at least* MinSize, while
+		// still re-using small buffers.
```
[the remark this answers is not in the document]
lthibault: That would result in undersized buffers being returned to `p.buckets[idx]`.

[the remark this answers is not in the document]
lthibault: I've reworked this a bit to make it more readable.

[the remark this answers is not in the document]
zenhack: But if we're returning oversized buffers anyway, shouldn't it be fine?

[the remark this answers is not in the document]
zenhack: Per discussion on matrix, for posterity: this is so that it's possible to return buffers not allocated with the Pool in the first place, so we'll keep it as-is.

exp/bufferpool/pool.go
```diff
@@ -1,52 +1,117 @@
...
+			return bs[i].Get().([]byte)
+		}
+	}
+
+	return nil
+}
+
+func (bs bucketSlice) Put(buf []byte) {
+	for i := range bs {
+		if 1<<i <= cap(buf) {
```
lthibault, line 97: Whoops, this is a bug.  Fixing.

zenhack (dismissed):

lthibault:
Ok, I think this is _finally_ ready to merge.  How do you feel about squashing this branch?

zenhack:
Once we figure out CI, SGTM.

PR #468, opened 2023-02-25
Pool capnp.Message

lthibault:
Merge after #466 & #467.

PR #470, opened 2023-02-25
Remove roSingleSegment arena

lthibault:
Stacks onto #468.  Merge that first.

This is a precursor to pooling segmets of `MultiSegmentArena`, as per discussion on Matrix.

Pooled messages obviate the need for message reuse in the `Decoder`, greatly simplifying its implementation.  Care has been taken to avoid an extra copy where `demuxArena` was formerly called.

## Benchmarks:

The performance degradation is due to MultiSegment arenas not drawn from the pool.  This will be fixed in a subsequent PR.

Tested on Go version 1.20

### Old

```
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkDecode-8   	     180	   5625283 ns/op	 170.66 MB/s	11193828 B/op	   40022 allocs/op
PASS
ok  	capnproto.org/go/capnp/v3	1.948s
```

### New

```
goos: darwin
goarch: amd64
pkg: capnproto.org/go/capnp/v3
cpu: Intel(R) Core(TM) i7-1068NG7 CPU @ 2.30GHz
BenchmarkDecode-8	171	13576764 ns/op	70.71 MB/s	12409647 B/op	50021 allocs/op
PASS
ok  	capnproto.org/go/capnp/v3	3.257s
```


lthibault:
I'm going to re-open this as another PR.

PR #471, opened 2023-02-27
Send promise resolution messages

zenhack:
This patch updates sendCap such that it checks whether it should send senderHosted or senderPromise for an export. If the latter, it also arranges for a Resolve message to be sent when the promise is resolved.

Still TODO:

- [x] We should look at incoming unimplemented messages and drop the resolved reference if the remote peer indicates it doesn't understand the Resolve (as is the case with go-capnp currently) -- otherwise we will leak caps.
- [x] We need to actually test this properly; it doesn't break existing tests, but also none of the new code is covered by them.
- [x] We should provide a decent API for creating and then fulfilling promises.
- [ ] We should start treating incoming senderPromise caps differently from senderHosted; we should rig up a promise for them, and then fulfill the promise when a Resolve comes in, rather than just ignoring it (responding unimplemented) as we do now.

Note: this patch adds a helper method to marshal an `error` into an `Exception`.


rpc/answer.go
```diff
@@ -312,8 +312,7 @@ func (ans *answer) prepareSendException(c *lockedConn, rl *releaseList, ex error
 			ans.c.er.ReportError(exc.WrapError("send exception", err))
 			ans.sendMsg = nil
 		} else {
-			e.SetType(rpccp.Exception_Type(exc.TypeOf(ex)))
-			if err := e.SetReason(ex.Error()); err != nil {
+			if err := e.MarshalError(ex); err != nil {
```
lthibault, line 315: maybe inline?  i.e.

```go
} else if err := e.MarshalError(ex); err != nil {

...

}

File: rpc/export.go [hunk 2]
```diff
@@ -154,26 +155,71 @@ func (c *lockedConn) sendCap(d rpccp.CapDescriptor, client capnp.Client) (_ expo
...
+		id = exportID(c.lk.exportID.next())
+		if int64(id) == int64(len(c.lk.exports)) {
+			c.lk.exports = append(c.lk.exports, ee)
+		} else {
+			c.lk.exports[id] = ee
+		}
+		c.setExportID(state.Metadata, id)
 	}
+	if ee.isPromise {
+		// Send a promise, wait for the resolution asynchronously, then send
```
lthibault, line 178: Could we abstract this block in a separate function?  `sendCap` is already quite long, and I think it would greatly improve readability.  I'd also be fine with just abstracting the anonymous goroutine into a separate (named) function.

rpc/export.go
```diff
@@ -154,26 +155,71 @@ func (c *lockedConn) sendCap(d rpccp.CapDescriptor, client capnp.Client) (_ expo
...
+		// a resolve message:
+		d.SetSenderPromise(uint32(id))
+		waitRef := client.AddRef()
+		go func() {
+			unlockedConn := (*Conn)(c)
```
lthibault, line 183: Do you want to assert that the conn is actually unlocked?  As I recall, you had a method that would panic if that wasn't the case.

No need if you're confident it's always unlocked.

lthibault (changes requested):

rpc/export.go
zenhack: I'm not sure what method your thinking of, or how it would even work.

In this case, this is at the top of a new goroutine, so it logically doesn't "own" c; when it needs the lock it should re-acquire it. It's a bit annoying that we can't get the unlocked version without a cast. Maybe this could use a comment.

zenhack: Update: added a comment.

rpc/answer.go
zenhack: Done.

rpc/export.go
```diff
@@ -154,26 +155,75 @@ func (c *lockedConn) sendCap(d rpccp.CapDescriptor, client capnp.Client) (_ expo
...
+		// a resolve message:
+		d.SetSenderPromise(uint32(id))
+		waitRef := client.AddRef()
+		go func() {
+			// Logically we don't hold the lock anymore; it's held by the
+			// goroutine that spawned this one. So cast back to an unlocked
+			// Conn before trying to use it again:
+			unlockedConn := (*Conn)(c)
+
+			waitErr := waitRef.Resolve(c.bgctx)
```
zenhack, line 188: Note: it might be better to use context.WithCancel and rig up the CancelFunc to be called if the entry is dropped. Otherwise there's a risk of leaking goroutines in a workflow where promises are created and never resolved.

rpc/export.go
zenhack: I agree in principle, though it's not obvious to me how to split this up in a way that's conceptually coherent; I'll have to think on it.

zenhack:
Added a (failing) test. Stay tuned.

lthibault (dismissed):

zenhack:
Test is fixed.

rpc/export.go
zenhack: Fixed this.

rpc/export.go
zenhack: Ok, split most of this out into a `sendSenderPromise` method.

zenhack:
Tacked on another commit to handle a remote peer that doesn't understand resolve messages; without this, we'd leak caps.

rpc/export.go
```diff
@@ -15,8 +16,12 @@ type exportID uint32
 
 // expent is an entry in a Conn's export table.
 type expent struct {
-	client   capnp.Client
-	wireRefs uint32
+	client    capnp.Client
+	wireRefs  uint32
```
lthibault, line 20: Just curious:  what are wireRefs?

lthibault (approved):
LGTM.  Note the failing CI though.

zenhack:
Yeah, I need to figure out why that test is failing -- I did some debugging, and the export is definitely being popped from  the table, so I'm not sure why Shutdown() isn't actually being called. I'll keep you posted.

rpc/export.go
zenhack: It's the protocol-level refcount.

zenhack:
Alright, last commit fixes the test.

lthibault (approved):
LGTM.

PR #473, opened 2023-03-01
Clean up rpc.Conn.receive

lthibault:
This is a refactor of `rpc.Conn` for readability.   There are no functional changes.

Clean-up is concentrated in two places:

1. The `switch` statement in `Conn.receive` is tidied up by having each `case` call out to a dedicated "handler" method.
2. <del>The `reader` thread is moved into its own task, allowing it to trigger task shutdown on its own.  This in turn lets us implement `receiver` as a simple `range` loop over a channel.</del>

A follow-up effort will focus on abstracting tasks & connection lifecycle.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -627,8 +627,7 @@ func (c *Conn) receive(ctx context.Context, incoming <-chan transport.IncomingMe
 			}
 		}
 
-		// incoming channel was closed; graceful exit.
-		return ctx.Err()
+		return nil
 	})
 }
 
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -255,26 +255,15 @@ func NewConn(t Transport, opts *Options) *Conn {
 		c.abortTimeout = 100 * time.Millisecond
 	}
 
-	// start background tasks
-	g.Go(c.backgroundTask(c.send))
-	g.Go(c.backgroundTask(c.receive))
-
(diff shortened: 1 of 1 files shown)


rpc/rpc.go
```diff
@@ -1755,28 +1755,59 @@ func (c *lockedConn) sendMessage(ctx context.Context, build func(rpccp.Message)
 // 'in' channel, until the context is canceled or an error occurs. The first
 // time RecvMessage() returns an error, its results will still be sent on the
 // channel.
-func reader(ctx context.Context, in chan<- incomingMessage, t transport.Transport) {
-	for {
-		inMsg, err := t.RecvMessage()
-		incoming := incomingMessage{
-			IncomingMessage: inMsg,
-			err:             err,
+func (c *Conn) reader(ctx context.Context, in chan<- transport.IncomingMessage) func() error {
```
zenhack, line 1758: I do not understand why or how this got more complicated; it seems like just wrapping the old implementation in backgroundTask should do the trick?

zenhack:
This broadly looks good to me, modulo the one comment about reader().

rpc/rpc.go
lthibault: Yeah, I woke up today and had the same though.  I've reverted this back to the original `incomingMessage` approach.

zenhack (approved):

PR #474, opened 2023-03-01
Remove unused method rpc.Conn.getAnswerQuestion.

lthibault:
This isn't used anywhere, despite what the comment says, so it seems like we could delete it.


zenhack:
#348, re-running.

zenhack (approved):
Good find.

PR #475, opened 2023-03-01
Don't export ClientPromise, use an interface instead.

zenhack:
I have at least one other type of fulfiller planned, and these are its only exported methods anyway.

In particular, I'm working on a `ClientHook` that wraps `AnswerQueue`, but I need to hook into the resolution logic to flush the queue.

### DIFF
--- a/fulfiller.go
+++ b/fulfiller.go
@@ -1,7 +1,7 @@
 package capnp
 
-// A Fulfiller supplies a value for a pending promise.
-type Fulfiller[T any] interface {
+// A Resolver supplies a value for a pending promise.
+type Resolver[T any] interface {
 	// Fulfill supplies the value for the corresponding
 	// Promise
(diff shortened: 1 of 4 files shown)


fulfiller.go
```diff
@@ -0,0 +1,12 @@
+package capnp
+
+// A Fulfiller supplies a value for a pending promise.
```
lthibault, line 3: Nit:  the name `Fulfiller` captures half of this type's functionality.  Ideally, we'd have a name that also captures the "reject" half.   

How about `Resolver`?

lthibault (changes requested):
LGTM, modulo bike-shedding on symbol names.

fulfiller.go
zenhack: Done

lthibault (approved):

PR #476, opened 2023-03-01
Add NewLocalPromise() api

zenhack:
Stacked on top of #475; merge that first.

This introduces a new function NewLocalPromise() which is similar to NewPromisedClient, but (1) is generic in the client type, so you can use it directly with anything that satisfies `~ClientKind`, and (2) doesn't take the `ClientHook` parameter, instead buffering calls in memory until the promise is fulfilled.

I haven't tested this, so I'm marking it as a draft.

I'm also not crazy about the names; I feel like we have too many distinct things called promise, and fewer of them should be exported anyway...

### DIFF
--- a/localpromise.go
+++ b/localpromise.go
@@ -11,10 +11,12 @@ type localPromise struct {
 	aq *AnswerQueue
 }
 
-func NewLocalPromise() (Client, Fulfiller[Client]) {
+// NewLocalPromise returns a client that will eventually resolve to a capability,
+// supplied via the fulfiller.
+func NewLocalPromise[C ~ClientKind]() (C, Fulfiller[C]) {
 	lp := newLocalPromise()
 	p, f := NewPromisedClient(lp)
-	return p, localFulfiller{
+	return C(p), localFulfiller[C]{
 		lp:              lp,
 		clientFulfiller: f,
 	}
@@ -52,17 +54,17 @@ func (lp localPromise) Reject(err error) {
 	lp.aq.Reject(err)
 }
 
-type localFulfiller struct {
+type localFulfiller[C ~ClientKind] struct {
 	lp              localPromise
 	clientFulfiller Fulfiller[Client]
 }
 
-func (lf localFulfiller) Fulfill(c Client) {
-	lf.lp.Fulfill(c)
-	lf.clientFulfiller.Fulfill(c)
+func (lf localFulfiller[C]) Fulfill(c C) {
+	lf.lp.Fulfill(Client(c))
+	lf.clientFulfiller.Fulfill(Client(c))
 }
 
-func (lf localFulfiller) Reject(err error) {
+func (lf localFulfiller[C]) Reject(err error) {
 	lf.lp.Reject(err)
 	lf.clientFulfiller.Reject(err)
 }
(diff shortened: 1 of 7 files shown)


localpromise.go
```diff
@@ -0,0 +1,70 @@
...
+	lp := newLocalPromise()
+	p, f := NewPromisedClient(lp)
+	return C(p), localFulfiller[C]{
+		lp:              lp,
+		clientFulfiller: f,
+	}
+}
+
+func newLocalPromise() localPromise {
+	return localPromise{NewAnswerQueue(Method{})}
```
lthibault, line 26: Nit:  linters often complain if you don't use keyed fields in struct literals.  Recommend changing to `localPromise{aq: NewAnswerQueue(Method{})}`.

lthibault (changes requested):

localpromise.go
zenhack: Done.

localpromise.go
```diff
@@ -0,0 +1,70 @@
...
+func (lp localPromise) Send(ctx context.Context, s Send) (*Answer, ReleaseFunc) {
+	return lp.aq.PipelineSend(ctx, nil, s)
+}
+
+func (lp localPromise) Recv(ctx context.Context, r Recv) PipelineCaller {
+	return lp.aq.PipelineRecv(ctx, nil, r)
+}
+
+func (lp localPromise) Brand() Brand {
+	return Brand{}
```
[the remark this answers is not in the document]
zenhack: We can always do so later if/when we need it.

lthibault (dismissed):

zenhack:
Added some tests, marking as ready.

zenhack:
Actually, these don't do exactly what I intended, back to draft for a moment.

rpc/localpromise_test.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+	"errors"
+	"testing"
+
+	"capnproto.org/go/capnp/v3"
+	"capnproto.org/go/capnp/v3/rpc/internal/testcapnp"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLocalPromiseFulfill(t *testing.T) {
```
lthibault, line 13: Can we set `t.Parallel()` ?

rpc/localpromise_test.go
```diff
@@ -0,0 +1,86 @@
...
+	res2, err := fut2.Struct()
+	assert.NoError(t, err)
+	assert.Equal(t, int64(2), res2.N())
+
+	res3, err := fut3.Struct()
+	assert.NoError(t, err)
+	assert.Equal(t, int64(3), res3.N())
+}
+
+func TestLocalPromiseReject(t *testing.T) {
```
lthibault, line 53: Ditto `t.Parallel()`

lthibault (changes requested):

rpc/localpromise_test.go
zenhack: Done.

zenhack:
Ok, did the additional testing I intended (checks order of receipt now).

lthibault (approved):

Issue #89, opened 2017-07-14
schemas: remove global registry


zombiezen:
Having a global registry of schemas creates problems with vendored copies of go-capnproto2, even if the application doesn't use reflection.  I'd like to change APIs that require reflection data to require it as a parameter.  I'm thinking that each `.capnp.go` should have a function that returns all the reflection data and bundles up the transitive dependencies.  This approach ought to lend itself to dead-code stripping, which would make binary sizes smaller.

PR #478, opened 2023-03-02
Remove more uses of fmt.

zenhack:
This patch:

- Drops a number of String() methods that require use of fmt, or importing the encoding/text subpackage.
- Avoids use of fmt in the schemas package.

After this patch, tempest's UI only pulls in fmt as a dependency of errgroup. I may look to see if we can avoid errgroup or build a lightweight alternative...

### DIFF
--- a/capnpc-go/fileparts.go
+++ b/capnpc-go/fileparts.go
@@ -21,7 +21,6 @@ var (
 		{path: flowcontrolImport, name: "fc"},
 
 		// stdlib imports
-		{path: "fmt", name: "fmt"},
 		{path: "context", name: "context"},
 		{path: "math", name: "math"},
 		{path: "strconv", name: "strconv"},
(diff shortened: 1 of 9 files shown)


zenhack:
Actually I think I see a way of doing this without dropping some of these; going to do another pass.

zenhack:
Ok, did another pass which should fix the tests; we now only get rid of the one String() method on StructList. Most others are left in place and instead we remove uses of fmt from encoding/text.

In addition to errgroup, the schemas package still transitively pulls in fmt via compress/zlib. #89 is related; we'd like to make embedding the schema at all optional.

zenhack:
Note also, this PR drops the size of tempest's WASM blob from 624K to 460K, so that's cool.

lthibault (approved):

PR #481, opened 2023-03-19
Fix order of comment numbers in test.

zenhack:
Probably I was shuffling these around and didn't update them when I was done.


zenhack:
#348, re-running.

lthibault (approved):

Issue #433, opened 2023-01-16
Improve Message.Reset API


lthibault:
It's not immediately obvious how to use `Message.Reset` to reuse a message from application code.  At a glance, I would expect to have to do the following:

1. `foo.Message().Reset(capnp.SingleSegment(nil))`
2. `foo, _ = NewRootFoo(foo.Segment())`, or some such

The "or some such" in item 2 says it all:  it is not immediately clear where the proper segment can be obtained.

An alternative suggested by @zenhack would be to use `NewMessage()`, though this will allocate an additional message.  Perhaps this can be mitigated with some clever use of `sync.Pool`?

zenhack:
Adding this to the 3.0 milestone, so we remember to consider it before locking ourselves in to the current API.

lthibault:
Relatedly, I would expect the following to work, but it produces a panic:

```go
type pointerStore struct {
	arena capnp.Arena
	msg   *capnp.Message
}

func (ps *pointerStore) Store(ptr capnp.Ptr) error {
	if ps.msg == nil {
		ps.msg = &capnp.Message{}
	}

	ps.arena = capnp.SingleSegment(nil)
	ps.msg.Reset(ps.arena)

	return ps.msg.SetRoot(ptr)
}
```

lthibault:
Another thought:  is it possible to have `Message.Reset()` call `Arena.Reset()` instead of having to call the latter separately, as we do in `transport.go`?

```go
	release := func() {
		if alreadyReleased {
			return
		}
		alreadyReleased = true

		msg.Reset(nil)
		arena.Release()  // fold this into msg.Reset ?
	}
```

zenhack:
Per discussion on matrix, we're going to call this one good; we'd both misunderstood what the API was for, and the docs have been clarified.

PR #482, opened 2023-03-20
reduce 'arena not empty' error allocation

piotrkowalczuk:
tl;dr: This PR preallocates `new message: arena not empty` error to improve performance. 

At this point I'm not sure yet whether the fact that `Decode` indirectly produces `new message: arena not empty` error is something expected, a bug or maybe a misuse on my end. Nevertheless it is a bit wasteful. CPU profile of a test that consumes over 1M messages looks like this:

image

And after the change:
image

### DIFF
--- a/message.go
+++ b/message.go
@@ -100,6 +100,8 @@ func NewMultiSegmentMessage(b [][]byte) (msg *Message, first *Segment) {
 	return msg, first
 }
 
+var errArenaNotEmpty = errors.New("new message: arena not empty")
+
 // Reset the message to use a different arena, allowing it
 // to be reused. This invalidates any existing pointers in
 // the Message, and releases all clients in the cap table,
@@ -127,11 +129,11 @@ func (m *Message) Reset(arena Arena) (first *Segment, err error) {
 				return nil, exc.WrapError("new message", err)
 			}
 			if len(first.data) > 0 {
-				return nil, errors.New("new message: arena not empty")
+				return nil, errArenaNotEmpty
 			}
 
 		default:
-			return nil, errors.New("new message: arena not empty")
+			return nil, errArenaNotEmpty
 		}
 
 		if first.ID() != 0 {


lthibault:
Hi @piotrkowalczuk!

First of all, thank you very much this PR and the analysis it contains.  This is a very nice catch!

>At this point I'm not sure yet whether the fact that Decode indirectly produces new message: arena not empty error is something expected

This is definitely not expected.  Do you have a sense of where/why it is being triggered?

I'm in favor of merging this change as a quick, low-cost fix, but I would also like to understand the root-cause.  (@zenhack?)

zenhack:
Hm, yeah I'd guess that if you're hitting an error like that on the hot path of your program something is wrong. Can you share an example of how you're using it?

piotrkowalczuk:
> This is definitely not expected. Do you have a sense of where/why it is being triggered?


`Message.Reset` at the very end of `Decode`. I'm not doing anything fancy, especially that `Decode` already takes care of many things. The code looks more or less like:

```go
// Read implements Scanner interface.
func (s *scanner) Scan() (capnp.Ptr, error) {
	msg, err := s.decoder.Decode()
	if err != nil {
		if errors.Is(err, io.EOF) {
			return capnp.Ptr{}, err
		}
		return capnp.Ptr{}, fmt.Errorf(".....: %w", err)
	}

	s.last = msg

	root, err := msg.Root()
	if err != nil {
		return capnp.Ptr{}, fmt.Errorf(".......: %w", err)
	}

	return root, nil
}

// Release implements Scanner interface.
func (s *scanner) Release() {
	if s.last != nil {
		s.decoder.ReleaseMessage(s.last)
	}
}


func example() {
	// ..... bunch of stuff   
	ptr, err := scanner.Scan()
	if err != nil {
		if errors.Is(err, io.EOF) {
			return err
		}
		return fmt.Errorf(".....: %w", err)
	}
	defer scanner.Release()
	// ..... bunch of stuff   
}
```

message.go
```diff
@@ -127,11 +129,11 @@ func (m *Message) Reset(arena Arena) (first *Segment, err error) {
 				return nil, exc.WrapError("new message", err)
 			}
 			if len(first.data) > 0 {
-				return nil, errors.New("new message: arena not empty")
```
piotrkowalczuk, line 130: This is the exact line where the issues occur.

zenhack:
I think you've found an actual bug in the Decode(); it's mis-using the Reset() api, doesn't check the error and it happens to "work."

Message.Reset is kindof an awkward API in the first place; it feels like you ought to be able to use it for reads like this but it doesn't technically support that. But I'm not sure why it couldn't; it seems like we could just remove the logic to reject existing data entirely and call it a day. @lthibault, thoughts?

lthibault:
SGTM.

zenhack:
@piotrkowalczuk, any interest in taking a whack at the fix to Reset I described?

zenhack:
As of #488 Decode() no longer uses Reset(), so this should be fixed. But I think the change to Reset would still make sense; maybe we should re-open #433 to track it?

PR #485, opened 2023-03-21
Simplify Client leak reporting.

zenhack:
This patch:

- Simplifies the metadata we collect; we have a full stack trace now, so the rest is redundant.
- Refactors so that dead code elimination can more easily drop the relevant code if SetClientLeakFunc is never called.
- Remove the metadata from the clients themselves, which will reduce the memory usage of clients by a bit when reporting is off.

I'm not entirely sure what moved me to work on this now, but here we are.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -117,11 +117,6 @@ type ClientKind = struct {
 }
 
 type client struct {
-	creatorFunc  int
-	creatorFile  string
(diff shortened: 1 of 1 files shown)


lthibault (approved):
Sometimes "no reason" is the best reason 😉 

LGTM!

PR #487, opened 2023-03-23
Move (de)coder to its own file

lthibault:
I thought I'd look into #482 for a bit, and see if I could pave the way for Piotr.  Along the way, I thought it might make sense to split up `message.go`, as it's getting quite long.

This PR splits `message.go` into two files:  `message.go` and `codec.go`, the latter containing `Encoder` and `Decorder`.  Ditto for `message_test.go`


zenhack:
Failure is #348; re-running.

PR #488, opened 2023-03-24
Pool multi-segment buffers.

lthibault:
I think this incidentally fixes #482 as well.

I recommend looking at commits individually.

### DIFF
--- a/arena.go
+++ b/arena.go
@@ -142,9 +142,10 @@ func MultiSegment(b [][]byte) *MultiSegmentArena {
 // will release the memory per usual.
 func (msa *MultiSegmentArena) Release() {
 	for i, v := range msa.ss {
+		msa.ss[i] = nil
+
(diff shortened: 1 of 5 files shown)


arena.go
```diff
@@ -160,75 +141,103 @@ func MultiSegment(b [][]byte) *MultiSegmentArena {
...
 // will release the memory per usual.
 func (msa *MultiSegmentArena) Release() {
-	for i, v := range *msa {
-		// Clear the memory, so there's no junk in here for the next use:
-		for j := range v {
-			v[j] = 0
+	for i, v := range msa.ss {
+		// segment not in buf?
+		if i >= msa.delim {
+			msa.ss[i] = nil
```
lthibault, line 147: TODO: nil out even if in buffer

zenhack (approved):

PR #489, opened 2023-03-24
Clean up Message.Reset

lthibault:


answerqueue.go
```diff
@@ -14,12 +14,12 @@ import (
 //
 // An AnswerQueue can be in one of three states:
 //
-//	1) Queueing.  Incoming method calls will be added to the queue.
-//	2) Draining, entered by calling Fulfill or Reject.  Queued method
-//	   calls will be delivered in sequence, and new incoming method calls
-//	   will block until the AnswerQueue enters the Drained state.
-//	3) Drained, entered once all queued methods have been delivered.
-//	   Incoming methods are passthrough.
+//  1. Queueing.  Incoming method calls will be added to the queue.
```
lthibault, line 17: I haven't made any changes here.  Must be `gofmt`.

rpc/transport/transport.go
```diff
@@ -56,14 +56,36 @@ type Transport interface {
 }
 
 type OutgoingMessage struct {
-	Message rpccp.Message
-	Send    func() error
-	Release capnp.ReleaseFunc
+	Message  rpccp.Message
+	send     func() error
```
zenhack, line 60: How will Transport implementations defined outside of this package set send?

rpc/transport/transport.go
```diff
@@ -56,14 +56,36 @@ type Transport interface {
...
+	}
+
+	panic("Tried to send() a message that was already released.")
 }
 
 type IncomingMessage struct {
-	Message rpccp.Message
-	Release capnp.ReleaseFunc
+	Message     rpccp.Message
+	ReleaseHook capnp.ReleaseFunc
```
zenhack, line 80: This field in particular could use some documentation; really not obvious what it does anymore. More broadly, we need to update the docs for Transport to reflect these changes.

zenhack (changes requested):
Mostly looks good, a couple issues.

rpc/transport/transport.go
lthibault: D'oh!  Good catch.  Will fix this ASAP.

lthibault:
@zenhack Comments addressed.  As per my comment in Matrix, I felt it was best to turn `OutgoingMessage` and `IncomingMessage` into interfaces, so the diff has grown to include their respective call sites.  Should be pretty straightforward though.

rpc/transport/transport.go
```diff
@@ -22,28 +22,31 @@ import (
...
 	//
-	// Messages returned by NewMessage must have a nil CapTable.
-	// When the returned ReleaseFunc is called, any clients in the message's
-	// CapTable will be released.
+	// OutgoingMessage instances returned by NewMessage must have a nil
+	// CapTable.  When OutgoingMessage.Release() is called, the underlying
+	// *capnp.Message will be released. This will also release any clients
+	// in the message's CapTable and release its Arena.
 	//
 	// The Arena in the returned message should be fast at allocating new
```
zenhack, line 35: Not really related to this PR (no need to address here) but: Is there actually a reason for this `should`? Afaik there's nothing in the RPC subsystem that would make SingleSegment a problem.

rpc/transport/transport.go
```diff
@@ -55,37 +58,32 @@ type Transport interface {
...
-
-type IncomingMessage struct {
-	Message     rpccp.Message
-	ReleaseHook capnp.ReleaseFunc
+// OutgoingMessage is a message that can be sent at a later time.
+// Release() MUST be called when the OutgoingMessage is no longer in
+// use. Before releasing an ougoing message, Send() MAY be called at
+// most once to send the message over the transport that produced it.
+//
+// Implementations MUST ensure the cap table is empty before Send is
```
zenhack, line 66: I think this MUST is incorrect; in fact, we keep the cap table full in the case of sending Return messages, in case we need to pipeline later questions on the result.

zenhack (changes requested):

rpc/transport/transport.go
lthibault: I always assumed it was because the `PlaceArgs` call might allocate lots of (potentially large) structs?  In other words,  it would be a case of "we're handing off control the user, so make sure we can handle whatever they do". 🤷‍♂️

zenhack (approved):

PR #491, opened 2023-03-27
Use go-util's mutex package for clientHook state.

zenhack:
Note: this depends on https://github.com/zenhack/go-util/pull/1.

I want to use this basically everywhere else we're using a mutex as well, but I to stay sane while reworking Client internals I wanted to do this now. I just did clientHook to keep this PR small-ish, but expect similar for related data structures soon.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -12,6 +12,7 @@ import (
 	"capnproto.org/go/capnp/v3/flowcontrol"
 	"capnproto.org/go/capnp/v3/internal/str"
 	"capnproto.org/go/capnp/v3/internal/syncutil"
+	"zenhack.net/go/util/sync/mutex"
 )
 
 func init() {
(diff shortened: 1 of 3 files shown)


capability.go
```diff
@@ -214,18 +224,19 @@ func (c Client) startCall() (hook ClientHook, resolved, released bool, finish fu
...
-	c.h = resolveHook(c.h)
+	l := c.h.state.Lock()
+	c.h, l = resolveHook(c.h, l)
 	if c.h == nil {
 		return nil, true, false, func() {}
 	}
-	c.h.calls++
-	c.h.mu.Unlock()
+	l.Value().calls++
+	isResolved := l.Value().isResolved()
```
zenhack, line 233: This here is the only functional change; previously isResolved was called in-line in the return statement, but this is incorrect according to isResolved()'s docs, which say you must be holding the lock. Making this change turned that into a type error, which brought it to my attention.

lthibault (approved):

lthibault (approved):

PR #493, opened 2023-03-28
Use go-util's mutex package in Promise.

zenhack:
Similar to #491/#492.

### DIFF
--- a/answer.go
+++ b/answer.go
@@ -4,11 +4,10 @@ import (
 	"context"
 	"errors"
 	"strconv"
-	"sync"
 
 	"capnproto.org/go/capnp/v3/exc"
 	"capnproto.org/go/capnp/v3/internal/str"
-	"capnproto.org/go/capnp/v3/internal/syncutil"
+	"zenhack.net/go/util/sync/mutex"
 )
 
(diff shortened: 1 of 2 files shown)


answer.go
```diff
@@ -34,11 +33,17 @@ type Promise struct {
...
 	//	- Resolved.  Fulfill or Reject has finished.
 
-	// mu protects the fields below.  When acquiring multiple Promise.mu
-	// mutexes, they must be acquired in traversal order (i.e. p, then
-	// p.next, then p.next.next).
-	mu sync.Mutex
+	// When acquiring multiple Promise.state mutexes, they must be acquired
+	// in traversal order (i.e. p, then p.next, then p.next.next).
+	//
+	// FIXME: is the above traversal scheme sound? It depends on avoiding
```
zenhack, line 39: Note: this is an existing problem, though I'm not sure whether we actually rely on this in the code anymore or if it's just the comment that's bitrotten -- I have vague memories of fixing a bug that was caused by this bad assumption, so we may be good.

Anyway, this is mostly unrelated, I just spotted it while making the change.

lthibault (approved):

PR #496, opened 2023-04-01
Simplify clientHook's refcounting scheme.

zenhack:
There is no actual reason why we need separate refcounts for calls and clients. In preparation for generally unifying the refcounting in this library, I've decided we should just merge them.

### DIFF
--- a/answer_test.go
+++ b/answer_test.go
@@ -103,7 +103,7 @@ func TestPromiseFulfill(t *testing.T) {
 		defer msg.Release()
 
 		res, _ := NewStruct(seg, ObjectSize{PointerCount: 3})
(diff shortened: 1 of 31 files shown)


lthibault (approved):
LGTM.  Note that I have not looked at CI failure.

zenhack:
I merged in #502, which fixed the test failures locally, so hopefully it should do so in CI as well. That PR should be reviewed & merged first.

zenhack:
#348, re-running.

lthibault (dismissed):
Approved (pending questions in #502)

lthibault (approved):

zenhack:
Per my long rambly rubber-ducking session on matrix, there *is* a reason for having two separate refcounts: if refs hits zero Client.Release() will block on calls hitting zero before returning. This patch breaks that, oops. Closing.

PR #497, opened 2023-04-01
Use go1.19 atomic.Uint64 in Message.

lthibault:
Minor cleanup in preparation for refcounted messages.  No functional changes.


lthibault:
@zenhack Note that `atomic.Uint64` is present in the standard library as of go 1.19, and that we dropped support for go1.18 a while back.

zenhack (approved):

PR #498, opened 2023-04-02
Don't directly export cap table from message

lthibault:
This PR paves the way for adding refcounting to `capnp.Message`.  Prior attempts to do so where frustrated by other packages accessing `Message.CapTable` directly.  This PR defines a `Message.CapTable()` method, that returns a new `CapTable` type.  The cap table is accessed and updated through the latter's methods.  This has the side benefit of simplifying code in quite a few places.

Note that this PR ended up rather large, but that most of the changes involve updating calls to `Message.AddCap` etc.  The diff is actually quite simple beyond that.  I recommend reviewing each PR separately, and focusing your attention on `message.go` and `captable.go`.


message.go
```diff
@@ -222,13 +212,8 @@ func (m *Message) SetRoot(p Ptr) error {
...
 }
 
-// AddCap appends a capability to the message's capability table and
-// returns its ID.  It "steals" c's reference: the Message will release
-// the client when calling Reset.
-func (m *Message) AddCap(c Client) CapabilityID {
-	n := CapabilityID(len(m.CapTable))
-	m.CapTable = append(m.CapTable, c)
-	return n
+func (m *Message) CapTable() *CapTable {
```
zenhack, line 215: This could probably use a brief doc comment.

rpc/rpc.go
```diff
@@ -1517,24 +1514,26 @@ func (c *lockedConn) recvPayload(rl *releaseList, payload rpccp.Payload) (_ capn
...
-		mtab[i], err = c.recvCap(ptab.At(i))
-		if err != nil {
-			for _, client := range mtab[:i] {
-				rl.Add(client.Release)
-			}
-			return capnp.Ptr{}, nil, rpcerr.Annotate(
-				err, "read payload: capability "+str.Itod(i),
-			)
+		if cl, err = c.recvCap(ptab.At(i)); err != nil {
+			cl.Release()
```
zenhack, line 1524: It is not safe to release clients while holding the connection lock; it could deadlock. This is why we passed in a releaseList.

zenhack (changes requested):
Mostly looks good, a couple comments in-line, on trivial and one critical.

zenhack (approved):

PR #499, opened 2023-04-10
Fix clerical errors in the comments for resolveHook

zenhack:


lthibault (approved):
Phew, that almost made it into production! 😉 😆

Issue #248, opened 2022-05-22
Add typed wrapper for `List[Client]`.


lthibault:
Now that we have support for mapping `:Capability` types to `capnp.Client`, it might make sense to add a `List[Client]` wrapper, along the same lines as #247.  Thoughts?

zenhack:
Makes sense to me.

zenhack:
Half-baked thought: what if we changed things so that instead of defining interface types as `type Foo struct { Client }`, we just did `type Foo Client`? Then, `CapList` could have its constraint set to `~Client`, and it would work for both `Client` itself and the typed wrappers.

We could probably extend this to other cases as well.

We would probably have to go back to generating some extra methods though, but that's maybe ok.

lthibault:
What does this get us?  Can we not set the constraint to `~struct{ Client }`? 

Note that I don't dislike this.  There are a lot of (nested) embedded structs in the library, and flattening that out ought to help with readability.

zenhack:
That's what we have now, but it means `CapList` can't be used for base `Client`, only the generated interface types that wrap it. This would allow it to be used for both.

lthibault:
Makes sense.  Go for it 👍

zenhack:
This is done, closing.

PR #501, opened 2023-04-11
handleCancel: minor simplification.

zenhack:
This should be more or less equivalent, but is less roundabout.


zenhack:
#248, re-running.

PR #503, opened 2023-04-13
flowcontrol/bbr: prune unused test functions

alrs:
This removes `Snapshot.report()`, `ceArg()`, and `computeErr()` from the tests in `flowcontrol/bbr`.


zenhack:
I'd actually rather not remove this; it's there basically because I left off in the midst of trying to hunt down some bugs, and will probably be useful when I get back to that, hopefully after I've finished implementing 3PH. Obviously at some point these should either be used or removed, but I'd like to punt that decision off until this package is working as expected.

Thanks anyway.

PR #506, opened 2023-04-19
exc.WrapError: return nil if the original error is nil.

zenhack:
I'm finding this convenient in Tempest


lthibault (approved):
I swear, I set out to make this change at least 5 times, and somehow never managed to open the friggin' PR.

Approved, obviously :)

PR #511, opened 2023-04-28
Client.peek(): make returns consistent with startCall()

zenhack:
The fact that these two were flipped is a regression waiting to happen.


lthibault (approved):
Oof, yeah. Nice one.

PR #515, opened 2023-05-11
Use the rc package for refcounting inside the implementation of Client.

zenhack:
This builds on top of #502, #508, and #509. There are still some failing tests, so I'm marking this as a draft, but it's close enough that it's probably worth a look.


lthibault (approved):

capability.go
```diff
@@ -127,6 +130,55 @@ type clientState struct {
...
+			return
+		}
+		l := res.Lock()
+		if !l.Value().isResolved() {
+			l.Unlock()
+			return
+		}
+		r := l.Value().resolvedHook
+		if r != nil {
+			r = r.AddRef()
```
lthibault, line 174: Why do we need the `AddRef` here?

capability.go
```diff
@@ -853,8 +819,8 @@ type ClientHook interface {
 
 	// Shutdown releases any resources associated with this capability.
 	// The behavior of calling any methods on the receiver after calling
-	// Shutdown is undefined.  It is expected for the ClientHook to reject
-	// any outstanding call futures.
+	// Shutdown is undefined. Shutdown must not interrupt any already
```
lthibault, line 822: Small nit/question:  is this a hand-wavy "must" or an RFC 2119 "MUST"?  I have code wherein long-running goroutines detect the canceled context upon shutdown, and abort.  That seems like it doesn't technically fall under "shutdown interrupted a call", but rather under "a call detects a shutdown and bows out", but I want to make sure this is okay.

If it is okay, then I think the docstring could use a bit of clarification.

lthibault (changes requested):
Okay, actually reviewed for real this time.  LGTM overall, but requesting a small clarification.

capability.go
zenhack: This is part of #502. See https://github.com/capnproto/go-capnp/pull/502#discussion_r1164318648. We could all-caps it I suppose.

capability.go
zenhack: The call to c.hook.Release() below will drop its reference to resolveHook, so we need to get our own reference here to make sure that doesn't spuriously release the hook.

capability.go
lthibault: Right, it did seem familiar!  No need to all-caps it, I was just referring to RFC 2119 to distinguish from "you probably won't need to bother with that".

It's being handled to my satisfaction in #502, so I think this is ready to merge. 👍

zenhack:
Trying to record outstanding issues:

- [ ] The semantics around weakrefs are different. This is actually a bit challenging to fix, and per discussion on matrix, it's not entirely clear that we care -- but at a minimum there is a test in the main package that will need fixing.
- [ ] There seems to be a new bug that results in leaking the clients created by a promise in recvCap, specifically for receiverAnswer.
- [ ] TestDuplicateBootstrap fails intermittently it seems the weakref in the import table entry is being dropped between returns, so the two clients end up with separate importClients, though they point to the same place. I'm not sure why this is happening; it's tempting to think it must be related to the other weakrefs thing, but I don't think the change in semantics actually explains it.

zenhack:
Fixed the client leak.

zenhack:
I went ahead and update the weakrefs test. I'm still a little stumped on TestDuplicateBootstrap, and am getting frustrated. I'm tempted to suggest merging this with the failure in place (maybe marking the test as flaky); the bug seems less serious than the one we're on track to fix, so maybe punt it until later.

capability.go
```diff
@@ -472,32 +494,48 @@ func (c Client) IsValid() bool {
...
 	for {
-		h, resolved, released := c.peek()
+		h, _, released := c.startCall()
+		defer h.Release()
 		if released {
 			return errors.New("cannot resolve released client")
 		}
 
-		if resolved {
+		if h == nil {
```
lthibault, line 529: Is this right?  I'd expect a nil `clientHook` to _not_ be resolved.

zenhack: I can't remember why I changed this. Pushed another commit that changes it back.

lthibault (approved):

PR #516, opened 2023-05-12
Promise: delete unnecessary releasedClients field

zenhack:
This was only used to make ReleaseClients idempotent -- but it is even without this, since looping over the nil clients will no-op anyway.


zenhack:
#348, re-running.

lthibault (approved):

PR #519, opened 2023-05-24
Clienthook snapshot

zenhack:
Only the last commit is relevant; this is stacked on top of many things.


zenhack:
(There are a few more commits now, all of which are relevant as I've merged main back in). There's a new failing test still that i need to pin down.

zenhack:
Fixed the test.

lthibault (approved):

PR #520, opened 2023-05-26
Use ClientSnapshot in the export table, rather than Client

zenhack:
This also tweaks the leak reporting to also check snapshots.


zenhack:
We will *probably* also need to do something with capnp.Promise/answers, but this is at least half of solving the 4-way race.

lthibault (approved):

zenhack:
I merged master back in, which should fix the CI failure. Test suite passes 1000 times in a row locally.

lthibault (approved):

PR #522, opened 2023-05-27
Add a Steal() method to Client and ClientSnapshot

zenhack:
...and use it to catch a bug; the test is failing, need to track it down.

### DIFF
--- a/capability.go
+++ b/capability.go
@@ -550,6 +550,13 @@ func (c Client) AddRef() Client {
 	})
 }
 
+// Steal steals the receiver, and returns a new client for the same capability
+// owned by the caller. This can be useful for tracking down ownership bugs.
+func (c Client) Steal() Client {
+	defer c.Release()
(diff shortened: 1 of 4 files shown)


zenhack:
Alright, fixed the failing tests.

zenhack:
Just noticed that the .AddRef() in localpromise.go actually introduces a leak... I'm somewhat inclined to punt for now though, because the way the code works *one* of those Fulfills wants to be borrowed, and #480 gets rid of one of them anyway, which will make this much easier to fix.

lthibault (approved):

lthibault:
Superseded by #530’s local-promise and incoming-resolution refactor. The two ownership consumers that motivated `Client.Steal` / the compensating `AddRef` no longer exist; `CapTable.Add` now directly consumes the fulfillment reference. The old patch no longer applies cleanly, and this is not a remaining 3PH prerequisite.

PR #525, opened 2023-05-27
Store ClientSnapshots in the CapTable.

zenhack:
Stacked on #522 and #524; review and merge those first. This makes the CapTable hang on to `ClientSnapshots` directly, which will be important for addressing the 4-way race in the context of answers.


zenhack:
Marking this as a draft; fussing with this on one of the follow-up features has convinced me that this is a bit error prone: you end up with two things to remember to release, and I'm hitting a cascade of leaks in branch for #480. I may rethink how to do this, time to sleep on it.

zenhack:
I think instead of doing this, I'm going to leave `CapTable` as just storing `Client`s, and try bringing back the `resultsCapTable` field that used to be stored in `rpc.answer`; it simplified things a bit to get rid of it, but I think bringing it back as a slice of snapshots will be much less intrusive than trying to stick snapshots in Message's cap table. Closing.

PR #536, opened 2023-06-30
Clean up the API for releaseExport

zenhack:
This includes two commits, the first of which is a trivial cleanup that came to my attention when working on the second, which is the real meat of this patch: instead of returning the snapshot and expecting the caller to release it, we take a deferred.Queue and schedule the release ourselves. This is less error prone, and I want to possibly attach some more resources to `expent` as part of the 3PH work, so this will also help with that.

There is no functional change.


lthibault (approved):
LGTM.

PR #537, opened 2023-06-30
Replace ErrorReporter with a structured logging interface.

zenhack:
The new interface:

- Has levels (debug, info, warn, error)
- Accepts arguments for structured logging

Per the comments, the interface is designed to be used with *slog.Logger, though other implementations are possible (and we already use one in test).


zenhack:
Note that there's some followup refactoring to do, such as getting rid of the use of ReportError() internally, and perhaps actually using the lower logging levels for non-fatal events.

lthibault (approved):

Issue #46, opened 2016-08-22
capnpc-go should not generate name collisions


zombiezen:
One of the notable examples is in persistent.capnp that the annotation and the type collide.  However, there are subtler cases, such as the `struct` field in schema.capnp should not collide with the `Struct` field of the wrapper struct.

zenhack:
Hope it's cool if I think out loud on this thread.

One obvious strategy: we need an injective function from legal capnproto idenfiers to legal go identifiers that don't collide with names that go-capnpc uses for helper functions (e.g. `*_ServerToClient`).

Facts we can potentially take advantage of:

* capnproto has different case rules for different kinds of identifiers, e.g. I think annotations are the only thing that can appear at top-level in a schema file that can start with a lower case letter. This *might* be be the only case where we have to worry about upper-casing a capnproto name causing a collision with another capnproto name on its own.
* underscores are legal in go names, but not in capnproto names. We already seem to use underscores as namespace seperators.
  * Idea: if a name would otherwise collide with a go or go-capnpc name, append a single underscore. e.g. `Struct` becomes `Struct_` 
* Filesystems on MacOS and Windows are case-insensitive by default, so if someone tries to name two files e.g. `web-session.capnp` and `webSession.capnp`, they're going to hit other problems anyway. So maybe we can just punt on this and say "don't do that." `$Go.name` still exists if someone gets screwed over by someone else's schema.
* There are some conventions that show up in sandstorm and the main repo re: what chars get used in filenames, so we can probably optimize at those, and for stuff that tends not to show up risk generating something ugly.

zenhack:
For characters in filenames that don't constitute legal go package names, we can replace them with strings that start with a double underscore, e.g. `c++.capnp` generates a package named `c__plus__plus`. Since dash (`-`) shows up often in idiomatic filenames, we can maybe special case this to camelCasing, taking advantage of the filesystem point I made above (so `websession.capnp` generates package `websession`, and `web-session.capnp` generates package `webSession`).

zenhack:
Obviously, a lot of these techniques are going to cause source-level breakage, don't know how you want to handle that.

zenhack:
We could also just make go_capnpc use names with a trailing underscore, that don't otherwise conflict with go keywords. This would let us introduce new ones without worrying about disturbing the names of exising schema.

zenhack:
[...]

PR #539, opened 2023-08-07
Annotations generate a name collision - fixed

davidhubbard:
This is a partial fix of https://github.com/capnproto/go-capnp/issues/46

> zombiezen commented on Aug 22, 2016:
>
> One of the notable examples is in persistent.capnp that the annotation and
> the type collide.

Also see https://github.com/capnproto/capnproto/issues/323

This commit changes the generated code for all annotations that start with a lowercase letter. Since the generated code in Go must start with an uppercase letter to be valid Go, this change appends an "_" to the name, regardless of whether there is a name collision already. (The thinking is that a name collision can happen in the future, and the renaming must be strictly an injective function, meaning that Go identifiers are always chosen the same regardless of other names. This way the codegen will not cause painful refactors in other code due to a name change later.)

The capnp files in this repo under `std` all have $Go.name() annotations due to this bug. This removes one of them (std/capnp/persistent.capnp) and replaces it with a unit test that compiles the generated code. Actually doing a "go build" is one step father than previous unit tests.

The thinking is to not add too many calls to "go build," since there is a lot of overhead and slow unit tests are discouraging. This commit appears to add about 250ms to the time it takes to run a "go test" in the capnpc-go subdir.


davidhubbard:
**Draft pull request**

`go test` in the capnpc-go dir still fails with: `./persistent-simple.capnp.out.go:49:33: undefined: context`

davidhubbard:
`level0_test.go:2270: log level error: rpc: incoming disembargo: send receiver loopback: send message: rpc: build message: target for receiver loopback does not point to the right connection (args: map[])    senderpromise_test.go:404: Received unexpected error: test.capnp:PingPong.echoNum: call on null client`

The current CI status already had that error. This PR doesn't address it, but at least it wasn't caused here. See [previous CI builds](https://github.com/capnproto/go-capnp/actions/runs/5443927637)

capnpc-go/capnpc-go_test.go
```diff
@@ -467,3 +472,144 @@ func nodeListString(n []*node) string {
...
+// * src/capnp/persistent.capnp
+// * Also found in this repo: std/capnp/persistent.capnp
+//
+// It contains two definitions:
+//   interface Persistent {}
+//   annotation persistent(interface, field) :Void;
+//
+// testdata/persistent-simple.capnp is a minimal reproducible test case for the
+// collision.
+func TestPersistent(t *testing.T) {
```
lthibault, line 514: If possible/reasonable, I would recommend running `t.Parallel()` here.  We have a fairly large test suite, and the parallelism really helps.

lthibault (changes requested):
This is just a first pass to pick out things that jump out.  Digging in more deeply now.

Looks great so far! 👍

lthibault (approved):
This is easily one of the best PRs I've ever seen!  Thanks for your contribution -- approved! 🎉

capnpc-go/capnpc-go_test.go
davidhubbard: Good idea, I added t.Parallle() at `capnpac-go_test.go:525`

lthibault (approved):

PR #540, opened 2023-08-07
Wrap errors in rpc.Conn.receive

lthibault:
This PR labels errors returned from message handlers in `rpc.Conn.receive`.  There is no functional change.  The intent is to facilitate debugging.

Note that this reverts #364, which may impact embedded applications wishing to reduce binary size.  At this point, I think we need to bias towards developer productivity, but I would like to keep this in mind and remove `fmt` at a later time.

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -4,6 +4,7 @@ package rpc // import "capnproto.org/go/capnp/v3/rpc"
 import (
 	"context"
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


rpc/rpc.go
```diff
@@ -206,10 +207,10 @@ type Options struct {
 // Logger is used for logging by the RPC system. Each method logs
 // messages at a different level, but otherwise has the same semantics:
 //
-// - Message is a human-readable description of the log event.
-// - Args is a sequenece of key, value pairs, where the keys must be strings
-//   and the values may be any type.
-// - The methods may not block for long periods of time.
+//   - Message is a human-readable description of the log event.
```
lthibault, line 210: This is triggered by `gofmt`.

davidhubbard (approved):
Approved, pretty straightforward

PR #541, opened 2023-08-08
Fix field names that collide with generated names #46

davidhubbard:
This is a partial fix of https://github.com/capnproto/go-capnp/issues/46

> zenhack commented on Sep 17, 2016:
>
> if a name would otherwise collide with a go or go-capnpc name, append
> a single underscore. e.g. `Struct` becomes `Struct_`

Also see https://github.com/capnproto/capnproto/issues/323

This commit changes the generated getter and setter methods for struct fields. If the field name would collide with code generated for any struct such as `Method()` or `String()`, the field getter becomes `Method_()` and the setter becomes `SetMethod_()`.

Technically only `Method_()` would be necessary, but who wants to maintain the mental state? If the getter changes, the setter also changing is just simpler conceptually.

The capnp files in this repo under `std` have $Go.name() annotations due to https://github.com/capnproto/go-capnp/issues/46. This removes one of them (std/capnp/compat/json.capnp) and updates the unit tests for code coverage testing purposes.


lthibault (approved):
LGTM 👍

PR #551, opened 2023-12-06
Slight improvement in installation docs

rexposadas:
Removed the use of dollar sign ($) so that the copy feature copies exactly what should go in the terminal. 

Before, what would be copied was:

`$ git clone https://github.com/capnproto/go-capnp`

After: 

`git clone https://github.com/capnproto/go-capnp`

### DIFF
--- a/docs/Installation.md
+++ b/docs/Installation.md
@@ -6,7 +6,7 @@ Then, run the following command to install the compiler plugin for the
 Go language:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lthibault:
Hmm, I'm a bit torn on this one.  The `$` prefix is often placed in README's to avoid blind copy-paste, but then again it seems to have caused you some confusion...

I guess we can err on the side of convenience, since we're not running any sensitive commands.

Oh, and thanks for the contribution!

lthibault (approved):

PR #553, opened 2024-02-13
Update README.md

spikebike:
This README update shows a work around that fixes the examples with newer versions of go, which mentions using 3.0 or the newest tagged release will fix the problem.

### DIFF
--- a/example/books/README.md
+++ b/example/books/README.md
@@ -12,3 +12,17 @@ cd ex1 && go build . && ./bookstest1
 
 cd ex2 && go build . && capnp encode ../books/books.capnp Book < ./book.txt | ./bookstest2 && cd ..
 ```
+
+If this results in:
+```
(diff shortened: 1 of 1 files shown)


lthibault:
Thanks for writing this up!  Out of curiosity, had you seen [this](https://github.com/capnproto/go-capnp/wiki/Getting-Started) when trying for the first time?  It's linked in the "Getting Started" section in the README.

I think we ought to clarify instructions either in the root-level README, or in the Wiki (preferred).  I'm not sure it's such a good idea to duplicate/fragment instructions in the examples.

lthibault:
After thinking about this some more, I'm closing this issue because go 1.18 isn't supported.  In fact, we should probably bump our minimum version to 1.21 at this point.

Thank you for submitting this nevertheless.  I really appreciate it.

PR #555, opened 2024-03-05
bufferpool: Switch to zeropool

matheusd:
This commit switches the bufferpool to use the zeropool implementation for sync pools.

The stdlib sync.Pool implementation has an issue where it causes an additional heap allocation per Put() call when used with byte slices.

github.com/colega/zeropool package has been specifically designed to work around this issue, which reduces GC pressure and improves performance.

This also fixes the bufferpool's pkg benchmark to use a new pool per test, to avoid other tests influencing the behavior of the benchmark and sets it to report the allocations.

### DIFF
--- a/exp/bufferpool/pool.go
+++ b/exp/bufferpool/pool.go
@@ -3,6 +3,8 @@ package bufferpool
 
 import (
 	"sync"
+
+	"github.com/colega/zeropool"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


lthibault (approved):
This is great!  I've been mildly annoyed with `sync.Pool`'s heap allocs for quite a while, so thank you for introducing me to this great little library.

Issue #554, opened 2024-02-27
Generate data with zero allocs


matheusd:
How can I generate/encode data with **zero** heap allocations? 

Example use case: I want to generate data to write into a file, so I need to allocate a single structure, message and arena. I can pre-allocate the arena buffer. But all my attempts at writing a loop fail to make it zero alloc (either the arena grows or message escapes to heap).

Here's the set of test benchmarks: https://github.com/matheusd/capnptest01/blob/master/bench_test.go

And here are my results:

```
BenchmarkSetText01-7     6966074               174.9 ns/op            99 B/op          0 allocs/op
BenchmarkSetText02-7     1000000              1100 ns/op             352 B/op          5 allocs/op
BenchmarkSetText03-7     2315973               517.9 ns/op            72 B/op          2 allocs/op
BenchmarkSetText04-7     2661026               429.9 ns/op           260 B/op          0 allocs/op
```

While benchmarks 01 and 04 are shown with 0 allocs, that's actually just the amortized growth of the arena which ends up becoming large in those tests.

lthibault:
If you call `Release()` on your message, the underlying buffers are returned to a `sync.Pool`.  This is the recommended way to avoid allocations.  I am not certain that there is a way to _guarantee_ zero-allocation (un)marshalling, though the `Arena` API is designed to support it in principle.

matheusd:
AFAICT, my `BenchmarkTest03()` was supposed to do that (through the use of `msg.Reset(arena)`) and indeed, the arena itself does not grow, but some objects (looks to be the `*Segment` created inside `Message.setSegment()` ) still escape to the heap.


```go

func BenchmarkSetText03(b *testing.B) {
	var msg capnp.Message
	arena := capnp.SingleSegment(nil)

	b.ReportAllocs()
	b.ResetTimer()

	for i := 0; i < b.N; i++ {
		seg, err := msg.Reset(arena)
		if err != nil {
			b.Fatal(err)
		}

		tx, err := NewTransaction(seg)
		if err != nil {
			b.Fatal(err)
		}

		err = tx.SetDescription("my own descr")
		if err != nil {
			b.Fatal(err)
		}
	}

	// b.Log(arena.String())
}
```

![profile001](https://github.com/capnproto/go-capnp/assets/1752493/33499a98-3224-4bea-9fc1-3c92c2981e1d)

matheusd:
[...]

PR #556, opened 2024-03-14
message: Fix reuse of first segment

matheusd:
This fixes the Message's Reset() call to allow reuse of the first segment.

Prior to this fix, the first segment was discarded after the first Reset call, effectively causing a new segment to be initialized on every Reset call.

By reusing the first segment, the number of heap allocations is reduced and therefore performance is increased in use cases where the message object is reused.

The fix involved associtating the segment to the message and fixing checks to ensure the data of the segment is re-allocated after the reset.

A benchmark is included to show the current performance of this.

Part of #554

### DIFF
--- a/message.go
+++ b/message.go
@@ -100,6 +100,11 @@ func (m *Message) Release() {
 func (m *Message) Reset(arena Arena) (first *Segment, err error) {
 	m.capTable.Reset()
 	for k := range m.segs {
+		// Optimization: keep the first segment so that the re-used
+		// Message does not have to allocate a new one.
+		if k == 0 && m.segs[k] == &m.firstSeg {
+			continue
+		}
 		delete(m.segs, k)
 	}
 
@@ -113,6 +118,7 @@ func (m *Message) Reset(arena Arena) (first *Segment, err error) {
 		DepthLimit:    m.DepthLimit,
 		capTable:      m.capTable,
 		segs:          m.segs,
+		firstSeg:      Segment{msg: m},
 	}
 
 	if arena != nil {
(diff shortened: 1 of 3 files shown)


lthibault:
Test failures seems unrelated.  Re-running.

Thank you for submitting this -- I am currently traveling, but will track this as closely as possible.

message.go
```diff
@@ -100,6 +101,9 @@ func (m *Message) Release() {
 func (m *Message) Reset(arena Arena) (first *Segment, err error) {
 	m.capTable.Reset()
 	for k := range m.segs {
+		if k == 0 && m.segs[k] == &m.firstSeg {
```
lthibault, line 104: A comment might be helpful here.  Something along the lines of:

```go
// optimization:  keep the first segment so that the re-used Message does not have to
// allocate a new one.
```

message.go
```diff
@@ -442,7 +447,7 @@ func alloc(s *Segment, sz Size) (*Segment, address, error) {
 		var err error
 		s, err = s.msg.allocSegment(sz)
 		if err != nil {
-			return nil, 0, err
+			return nil, 0, fmt.Errorf("allocSegment failed: %v", err)
```
lthibault, line 450: We previously purged `fmt` form the code-base in order to avoid using reflection, which helps keep binary sizes small.  I know it's a bit awkward, but can we replace this with `errors.New("allocSegment failed: " + err.Error())`?

segment.go
```diff
@@ -15,6 +15,8 @@ type SegmentID uint32
 // It is part of a Message, which can contain other segments that
 // reference each other.
 type Segment struct {
+	// msg associated with this segment. A Message instance m maintains the
```
lthibault, line 18: Very helpful, ty!

lthibault (changes requested):
Looks great overall.  Small changes requested inline.

message.go
matheusd: Ah, I wasn't aware of it. It would be an improvement (IMO) then to create specific errors for the possible error causes and then manually wrap them (e.g. similar to how https://github.com/decred/dcrd/blob/master/internal/blockchain/error.go has `ErrorKind` and `ContextError`) which would remove the reliance on `fmt` and allow a finer grained check for errors.

But that's a project for the future... :P

lthibault (approved):

PR #559, opened 2024-04-04
docs: Add tip how to get to capnproto dir with go list

karelbilek:
I spent some time figuring out how to compile capnpn files in a relatively portable way; this seems to be the best way.


lthibault (approved):
This is great, ty!

PR #560, opened 2024-04-05
Refactor

matheusd:
Requires #555, #556

Elided tests and creating as draft to get a first pass review.

The diff is large(ish) so it may be easier to read the full code vs the diff at the moment. But the basic idea of the refactor is the following:

#### Split allocator strategy and segment management

Using the bufferpool should not be forced upon the caller. And in fact, it is dangerous to use the current implementation of single/multi segment arenas if you send a buffer allocated from anywhere _not_ there. For example, an mmaped file buffer.

Spinning the allocator into its own thing means we can define different allocation strategies (bufferpool, regular runtime functions, simpler caching strategy, read-only, etc).

Unfortunately, due to some tests failing otherwise, I couldn't unify literally everything inside the allocator (see further below for discussion).

#### Base arena implementation

SingleSegment and MultiSegment arenas have been unified into a single arena impl that offloads the logic to the allocator and segment list.

So the full matrix of Single/Multi and BufferPool/runtime-backed options can be exercised.

#### Reduce message complexiy

Most decisions have been offloaded from message into the arena, segment list or allocator. This makes `Message` more generic and easier to reason about: in particular, it no longer cares how many segments there are during init (see further below for discussion on Release) and the roundabound way it used to initialize the first segment.

#### Test compatibility and issues

All the existing tests pass. A few that are no longer applicable (dealing with the concrete arena implementations) have been commented out. One test (`TestPromiseOrdering`) is skipped because it's flaky even in the current main branch.

Other than those, the code has been specifically designed to not require changes in the existing tests, and therefore should be ensuring full compatibility to the existing code.

Tests for the new features haven't been done yet (but will if this is deemed to be in the right direction).

#### Message.Release is full of special cases

One main source of frustration during this rewrite is that `Message.Release` is full of special cases, mostly to deal with initializing a message for writing. It has all these cases I had to add to avoid having to touch the existing tests. These special cases are documented now in the code, after a `FIXME(matheusd)` line in that function.

Personally, I think my `ReleaseForRead()` should be the actual implementation of `Release`, but in the interest of not breaking client code, I opted for adding a new function instead.

This is also somewhat the reason for having to add a `ReadOnlySingleSegmentArena` instead of using a read only allocator: Release() is (currently) expected to check the arena is clear and re-allocate the first segment (i.e. "Prove Reset() cannot be used to reset a read-only message" commit), so I had to go out of my way to create an arena that would make it easier to just read only, while reusing the message struct.

### DIFF
--- a/allocator.go
+++ b/allocator.go
@@ -1,0 +1,275 @@
+package capnp
+
+import (
+	"errors"
+
+	"capnproto.org/go/capnp/v3/exp/bufferpool"
+)
+
+// allocator defines methods for an allocator for the basic arena
+// implementation: the allocator defines the strategy of how to grow and
+// release memory slices that back segments.
+type allocator interface {
+	// Grow an array. When called, it has already been determined that
+	// the slice must be grown. Implementations must copy the contents of
+	// the existing byte slice to the new one.
+	//
+	// The returned slice may be have a capacity larger than strcictly
+	// required to store an addition sz bytes. In this case, the length
+	// must be len(b)+sz, while cap may be any amount >= than that.
+	Grow(b []byte, totalMsgSize int64, sz Size) ([]byte, error)
+
+	// Release signals that the passed byte slice won't be used by the
+	// arena anymore.
+	Release(b []byte)
+}
+
+// bufferPoolAllocator allocates buffers from a buffer pool. If nil, then
+// the global default buffer pool is used.
+type bufferPoolAllocator bufferpool.Pool
+
+func (bpa *bufferPoolAllocator) Grow(b []byte, totalMsgSize int64, sz Size) ([]byte, error) {
+	pool := (*bufferpool.Pool)(bpa)
+	if pool == nil {
+		pool = &bufferpool.Default
+	}
+
+	inc, err := nextAlloc(totalMsgSize, int64(maxAllocSize()), sz)
+	if err != nil {
+		return nil, err
+	}
+	nb := pool.Get(len(b) + int(inc))
+
+	// When there was a prior buffer, copy the contents to the new buffer,
+	// clear the old buffer and return the old buffer to the pool to be
+	// reused.
+	if b != nil {
+		copy(nb[:cap(nb)], b[:cap(b)])
+		for i := range b {
+			b[i] = 0
+		}
+		pool.Put(b)
+	}
+
+	return nb, nil
+}
+
+func (bpa *bufferPoolAllocator) Release(b []byte) {
+	if b == nil {
+		panic("nil buffer passed to release")
+	}
+	pool := (*bufferpool.Pool)(bpa)
+	if pool == nil {
+		pool = &bufferpool.Default
+	}
+	pool.Put(b)
+}
+
+// simpleAllocator allocates buffers without any caching or reuse, using the
+// standard memory management functions.
+//
+// This allocator is concurent safe for use across multiple arenas.
+type simpleAllocator struct{}
+
+func (_ simpleAllocator) Grow(b []byte, totalMsgSize int64, sz Size) ([]byte, error) {
+	inc, err := nextAlloc(totalMsgSize, int64(maxAllocSize()), sz)
+	if err != nil {
+		return nil, err
+	}
+	return append(b, make([]byte, inc)...), nil
+}
+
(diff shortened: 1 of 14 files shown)


matheusd:
[Here](https://github.com/capnproto/go-capnp/pull/560/commits/1651df3b9f5ec3b3aa69d6dbe27a573c884688a4) is another find.

`NewStruct()` followed by `SetRoot()` is pretty inefficient. All the struct copying going around the calls are making them somewhat heavy. Unrolling the call to allocate and set the root pointer in a single function more than triples the performance in CPU terms:

```
BenchmarkMessageSetRoot-7          3173535      380.7 ns/op    0 B/op     0 allocs/op
BenchmarkMessageAllocateAsRoot-7   9872070      121.0 ns/op    0 B/op     0 allocs/op
```

Additionally, this makes it much easier to reason about what "setting the root of the message" entails and the code is much flatter.

Also, this gets rid of the need to have the root pointer allocated on `Reset()`, moving the message to an "allocate on first set" paradigm, getting rid of some of the quirks in `Reset()`.

The new `AllocateAsRoot()` can easily replace the existing code by modifying the generator to use that for the `NewRootXXXX` calls.

matheusd:
For the [next update](https://github.com/capnproto/go-capnp/pull/560/commits/c93652fe0c3c04d47202fd3ac92b3805c4a4691f) I introduce an unrolled version of `SetNewText`, which gets rid of about 100ns of cpu time for setting a text in a struct.

The `SimpleSingleSegmentArena` is there as a simpler implementation of the single segment arena that gets rid of all the indirection and caching for allocation, so when it has a correctly sized buffer it ends up faster than the standard arena impl.

Bringing everything so far, we get the following benchmark results, which we can compare to the ones in #554:

```
(Earlier attempts)
BenchmarkSetText01-7     6966074               174.9 ns/op            99 B/op          0 allocs/op
BenchmarkSetText02-7     1000000                1100 ns/op           352 B/op          5 allocs/op
BenchmarkSetText03-7     2315973               517.9 ns/op            72 B/op          2 allocs/op
BenchmarkSetText04-7     2661026               429.9 ns/op           260 B/op          0 allocs/op

(With the refactor)
BenchmarkSetTextFlat-7  26186197               50.69 ns/op             0 B/op          0 allocs/op
```

matheusd:
And for the [next update](https://github.com/capnproto/go-capnp/pull/560/commits/bde3071c91a90ec81fbaf1dd12226b8004c48e6a), I introduce an unrolled `UpdateText` version which reuses the storage for a string instead of always allocating a new one. This removes the need to reset the arena if you're rewriting a single field in a large structure.

This _significantly_ reduces the latency for an operation that consists only of updating a field, down to only ~16ns:

```
BenchmarkSetTextUpdate
BenchmarkSetTextUpdate-7        80923449                15.90 ns/op            0 B/op          0 allocs/op
```

lthibault:
@matheusd Thank you so much for all this incredible work!  I'll review this asap and make sure it gets the attention it deserves.  Probably won't be today, but know that you're on my radar! 🙂 🙏

matheusd:
Sure thing. I still have at least one additional idea to test out to make this particular workflow faster.

Also note that these are all wip, and specially the later pushes are all experimental stuff that I wouldn't expect to merge as is, but rather to use as a baseline to refactor the code to reach these benchmark results.

matheusd:
Pushed a [new experiment](https://github.com/capnproto/go-capnp/pull/560/commits/e7d8d6ffc62f046a0b7bc7084097f360df603a6e) that adds a `TextField` definition, which keeps track of the pointer and value locations of a specific field (basically a slimmed down `Struct`).

By keeping these around, we can forgo the most costly op from `UpdateText` (which is the `resolveFarPointer` call) verify directly whether we can copy the new string, then replace it.

This gets to within 2.2x of a baseline benchmark that just copies the data into a slice:

```
BenchmarkSetTextAsField-7       281488827                4.248 ns/op           0 B/op          0 allocs/op
```

lthibault:
@matheusd I'm getting ready to dive into this, and firstly wanted to thank you again for the detailed overview.

Now that you seem to be converging on an implementation, I'm wondering if there's any part of this that we can break off into a smaller PR and merge separately?

PR #565, opened 2024-05-19
Allow users to specify the Arena implementation used by server.

lthibault:
Motivation:  I want to use a custom `Arena` implementation in [Wetware](https://github.com/wetware/go).


lthibault:
@davidhubbard I am quite confident these are innocuous changes and the `server` package tests succeed, so I think it's safe to proceed despite test failures (slowly working up the courage to address those).

davidhubbard (approved):

PR #578, opened 2024-07-31
bbr: Fix test flakes

matheusd:
This fixes test flakes that happen due to timing of clock actions.

The tests involve multiple goroutines that access the current time of the clock, but it may happen that an Advance() call in the test happens before a read of the clock time to create a new timer, thus causing a test flake.

This issue is fixed by adding appropriate synchronization.


lthibault (approved):
Great find. TY for the fix!  This will be a crazy huge help when diving back into flow-control code.

PR #582, opened 2024-08-07
Add workarounds to TestPromiseOrder

matheusd:
These are merely workarounds and not a full solution to the underlying issues, but they (hopefully) allow the CI to stop failing spuriously and allow us to continue developing and improving the library with greater confidence on the test's reliability.


matheusd:
For completeness, I should add that an alternative is to disable (via `t.Skip()`) the `TestPromiseOrder` test.

lthibault (approved):
LGTM 👍

PR #584, opened 2024-08-13
CI: Reduce number of test runs of rpc

matheusd:
This reduces the number of sequential tests of the RPC subsystem from
500 to 50. This speeds up the run of the CI and reduces flakiness.


lthibault:
LGTM.

A note for posterity:  this was originally set to 500 to track down some rare race conditions.  I'm pretty confident it's no longer needed, especially with your most recent changes.

lthibault (approved):

PR #589, opened 2024-08-21
Simplify TestCanonicalize

matheusd:
This changes TestCanonicalize to use a table-driven approach and
individual t.Run() calls for each subtest.

This makes it easier to execute individual (sub-)tests and to trace any
errors. It also makes the test code easier to understand.

### DIFF
--- a/canonical_test.go
+++ b/canonical_test.go
@@ -1,161 +1,126 @@
 package capnp
 
 import (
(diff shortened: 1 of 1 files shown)


lthibault (approved):
Yeah, way better!  Thank you!

PR #590, opened 2024-08-28
Allow using the packed transport in rpc.Serve

fpetkovski:
The Serve function in the rpc module can only use the basic streaming encoding.

This commit allows setting the transport by the user through the appropriate ServeOption.


matheusd (approved):

lthibault (approved):
Nice!  Thank you!

PR #591, opened 2024-08-30
multi: Make NewMessage() usable for creating messages for reading

matheusd:
This PR modifies `Reset()` (and consequently `NewMessage()`) so that it can be used to initialize messages for reading.

Before this PR, `Reset()` would enforce that space for the root pointer was allocated in the associated arena, which could not be done for read-only messages. This PR changes it so that space for the root pointer is only allocated when strictly necessary. 

In the future, this will allow protecting Message values from wrong usage by enforcing the use of NewMessage to initialize message objects (instead of use of zero valued messages).

> [!WARNING]  
> This is a breaking API change. `NewMessage()` would fail if the underlying arena could not allocate space, and if this PR is merged that will not be true anymore.

### DIFF
--- a/answer_test.go
+++ b/answer_test.go
@@ -59,7 +59,7 @@ func TestPromiseFulfill(t *testing.T) {
 	t.Run("Done", func(t *testing.T) {
 		p := NewPromise(dummyMethod, dummyPipelineCaller{}, nil)
 		done := p.Answer().Done()
-		msg, seg, _ := NewMessage(SingleSegment(nil))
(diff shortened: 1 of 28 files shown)


lthibault (approved):
Hey @matheusd, my sincerest apologies for letting this sit for so long.  I've been dealing with some rather difficult life circumstances the past few months, but I am progressively easing back into things.

These changes LGTM.  Can you remind me where we are in your roadmap?  Is there any follow-up in the pipeline?

PR #594, opened 2024-10-12
3PH part 1: Prepare for 3PH support

Semisol:
This PR does not implement any 3PH related functionality, except preparations:
- A new `Network3PH` interface has been added to be implemented by networks supporting 3PH. (this could be named better)
- 3PH-related methods have been moved out of `Network` for networks that do not support it.
- Schemas have been updated to match the capnproto/capnproto repository.

To be done:
- [ ] Initial 3PH support, with forwarding and `Resolve` with 3rd party capabilities
- [ ] Embargo support for 3PH
- [ ] An extension to the API that allows tail calls (This is not going to be a priority for me, as it is only useful in rare conditions)


lthibault:
I think there's a dependency issue causing CI to fail.  I had a look in the meantime, and it LGTM overall.  Let's get CI passing and then merge it in.

lthibault (approved):

PR #595, opened 2024-11-28
Provide an option to specify base context for RPC connection

homier:
## Rationale

Providing a function to build base context is really useful in situations, when we need to propagate some, let's say, identity data to server handlers. Let's take the following issue as an example:
- "I need to pass the remote peer ID to my underlying handlers, so I could do some basic identification stuff"

The problem is that now the only way to do this is to set it in `rpc.Conn` and pass the connection to handlers, but this breaks abstraction between transport layer and application layer.
But we could do a simple change to have a context factory, so all the necessary connection information could be set in context, that will be passed to handlers later. And it's similar to how the native `http.Server` interface is [built](https://pkg.go.dev/net/http#Server).

## Real-life example (using libp2p as network stack):
```go
package handler

import 	(
       "capnproto.org/go/capnp/v3/rpc"
       "github.com/libp2p/go-libp2p/core"
)

type Handler struct {
    ...
}

func (h *Handler) Serve(stream core.Stream) *rpc.Conn {
        remotePeerID := stream.Conn().RemotePeer()

	conn := rpc.NewConn(rpc.NewPackedStreamTransport(stream), &rpc.Options{
		BootstrapClient: h.resolverHandler.Client(),
		BaseContext: func() context.Context {
		       return context.WithValue(context.Background(), "peerID", remotePeerID)
		},
	})

	return conn
}
```

### DIFF
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -91,7 +91,7 @@ type Conn struct {
 	baseContext  func() context.Context
 
 	// bgctx is a Context that is canceled when shutdown starts. Note
-	// that if baseContext is not provided. it's parent is context.Background(),
+	// that if baseContext is not provided, it's parent is context.Background(),
 	// so we can rely on this being the *only* time it will be canceled.
 	bgctx context.Context
 
@@ -204,7 +204,7 @@ type Options struct {
 	// set this.
 	Network Network
 
-	// BaseContext is an optional funcation that returns a base context
+	// BaseContext is an optional function that returns a base context
 	// for any incoming connection. If ommitted, the context.Background()
 	// will be used instead.
 	BaseContext func() context.Context
--- a/rpc/rpc.go
+++ b/rpc/rpc.go
@@ -88,10 +88,11 @@ type Conn struct {
 	bootstrap    capnp.Client
 	er           errReporter
 	abortTimeout time.Duration
+	baseContext  func() context.Context
 
 	// bgctx is a Context that is canceled when shutdown starts. Note
-	// that it's parent is context.Background(), so we can rely on this
-	// being the *only* time it will be canceled.
+	// that if baseContext is not provided. it's parent is context.Background(),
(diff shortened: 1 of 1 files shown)


fpetkovski (approved):

homier:
@lthibault Hi! Don't know whether it's ok to ping you, but could you review the PR please, if/when you have some free time? :) Thanks in advance!

lthibault:
hey @homier, yes, totally okay to ping me!  Reviewing now :)

lthibault (changes requested):
Firstly, thank you for the contribution!  Just a quick question before I approve:  have you tested that the behavior is the one you expect when the context is canceled?  I'd expect it to close the connection, and judging from the code, that _should_ happen.

Can we add a quick test for this behavior? 

P.S.:  as a libp2p developer myself, I'm very curious to know what you're building :)

homier:
@lthibault Thanks for the review! I'll add a test as soon as possible when I have some free time :) Expect a new ping from me 🙈

>  Just a quick question before I approve: have you tested that the behavior is the one you expect when the context is canceled? I'd expect it to close the connection, and judging from the code, that _should_ happen.

Yeah, everything works fine, I'm using my fork with this patch at the moment, works as expected. But anyway, I'll add some tests to ensure that expected behaviour is not broken.

> I'm very curious to know what you're building :)
Nothing special, I'm just doing my pet project with p2p file sharing in a form of mobile application at the moment. The code is closed now, though I'm planning to open it when I won't be ashamed of its quality :D

lthibault:
>Yeah, everything works fine, I'm using my fork with this patch at the moment, works as expected. But anyway, I'll add some tests to ensure that expected behaviour is not broken.

I'll merge in the meantime, just for the sake of expediency.  Just go ahead and add the test in a separate PR when you get a chance.

Thanks again for your contribution!

lthibault (approved):

PR #598, opened 2025-01-07
MultiSegmentArena: Fix release back into bufferpool

matheusd:
This fixes a bug that was preventing correct release of buffers allocated by Decode() back to the global buffer pool.

Previously, the data variable was overriden on the loop that splits the segments, preventing the correct tracking of the full data buffer and its subsequent return back to the buffer pool when the arena was released.

This also adds a .Release() call to the Decode benchmark to better reflect the ability to reuse buffers.

### DIFF
--- a/arena.go
+++ b/arena.go
@@ -267,6 +267,7 @@ func (msa *MultiSegmentArena) demux(hdr streamHeader, data []byte, bp *bufferpoo
 		msa.segs = append(msa.segs, make([]Segment, inc)...)
 	}
 
+	rawData := data
 	for i := SegmentID(0); i <= maxSeg; i++ {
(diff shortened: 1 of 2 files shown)


fpetkovski (approved):

lthibault:
@fpetkovski Just want to say "thank you" for your reviews. 🙏 

@matheusd Aiming to have this reviewed in the next few days.

matheusd:
Cool, thanks!

I have another small improvement that depends on this, to be pushed after this one is merged.

lthibault (approved):
Thanks for the benchmarks @Semisol.  This is a big perf win!  I feel like we should be communicating publicly about this kind of improvement.  Great work @matheusd!

PR #605, opened 2025-02-19
Fix broken AnswerQueue behavior with chained pipeline calls

### LINKED ISSUE #521: pipelined calls sometimes return unimplemented
Sometimes making a pipelined call on a result will spuriously return unimplemented, without invoking the actual object. @lthibault and @aratz-lasa report that they've seen this from time to time, and it seems to be the root of zenhack/tempest#11. Waiting on the result seems to work around the problem consistently.

### LINKED ISSUE #604: Pipelining in client on certain situations triggers "unknown method" in the "parent" capability
When you start a pipeline request from the client, while another request is in-flight, go-capnp won't recognise one of the RPC calls that comes in that were pipelined.
 
This is an issue we encountered and it was a bit hard to repro, here is a repo where you can easily reproduce (rust capnp client, but we also repro'd with a c++ one). 
https://github.com/gabivlj/capnp-issue

Schema:
```

interface Service {
    get @0 () -> (bsr :ByteStreamReturner);
}

interface ByteStreamReturner {
    getConnector @0 () -> (conn :Connector);
    inflighter @1 ();
}

interface Connector {
    connect @0 (down: ByteStream) -> (up :ByteStream);
}

interface ByteStream {
  write @0 (bytes :Data) -> stream;
}

```

Important part of the repro:
```

func (c *byteStreamGetter) GetConnector(ctx context.Context, getParams bytestream.ByteStreamReturner_getConnector) (rerr error) {
	getParams.Go()
	fmt.Println("Get from byteStreamGetter")
	res, err := getParams.AllocResults()
	if err != nil {
		return err
	}

	bServer := bytestream.Connector_NewServer(&conn{})
	bServer.HandleUnknownMethod = func(m capnp.Method) *server.Method {
		fmt.Println("Unknown method:", m.String())
		// So here, if you for example wait for the bytestream
		// you return in Connect and do something like this:

		/*
			        b := <-p.c
				fmt.Println("Got bytestream, returning impl")
				methods := bytestream.ByteStream_Methods([]server.Method{}, b)
				s := &server.Method{methods[m.MethodID].Method, methods[m.MethodID].Impl}
				p.c <- b
		*/

		// It will work!
		//
		// IDK why capnp-go here doesn't really understand what's going on
		return nil
	}

	// The more you sleep, the easier to repro it is. Not even sleeping
	// and you will repro this.
	time.Sleep(time.Millisecond)
	client := capnp.NewClient(bServer)
	b := bytestream.Connector(client)
	return res.SetConn(b)
}

type conn struct {
}

func (c *conn) Connect(ctx context.Context, params bytestream.Connector_connect) error {
	res, err := params.AllocResults()
	if err != nil {
		return err
	}

	bServer := bytestream.ByteStream_NewServer(&byteStreamWrite{})
	bServer.HandleUnknownMethod = func(m capnp.Method) *server.Method {
		panic("this one is never triggered")
		return nil
	}

	client := capnp.NewClient(bServer)
	b := bytestream.ByteStream(client)
	return res.SetUp(b)
}

type byteStreamWrite struct {
}

var a atomic.Int64

func (c *byteStreamWrite) Shutdown() {
	log.Println("ByteStream Shutdown()")
}

// Write writes the bytes to the connection.
func (c *byteStreamWrite) Write(ctx context.Context, params bytestream.ByteStream_write) error {
	a.Add(1)
	params.Go()
	fmt.Println("Received Write", a.Load())
	args := params.Args()
	bytes, err := args.Bytes()
	if err != nil {
		return err
	}

	fmt.Println("Write called (good!)", string(bytes))
	return nil
}
```

You will see the output:
```
~/repro$ bin/repro
Inflighter BS start
Get from byteStreamGetter
Unknown method: @0xb45bc9c79a817b99.@0
Inflighter BS done
2025/01/31 03:32:07 CAPNP ERROR: reader: transport: stream transport: receive: EOF
2025/01/31 03:32:07 conn done, accepting another
byte stream getter Shutdown()
2025/01/31 03:32:07 ByteStream Shutdown()
```

What we are doing in the client is basically just create a pipeline that does getConnector() -> connect() -> write().
The unrecognised method is `write`, and is triggered in the capability returned by getConnector(), so it makes me think go-capnp is somehow confused and thinks `write` belongs to connector, when it belongs to bytestream.

If you need any more reproduction steps or more clarification, please let me know.

ObsidianMinor:
AnswerQueue does not handle chained pipeline calls correctly. I'm assuming this is because the code for it might be incomplete. It doesn't assign basis recv functions for any bases other than 0 and uses the wrong value for the basis itself on subsequent chained calls.

This fixes those bugs and adds more documentation about how AnswerQueue is meant to function because I had to stare at it for 2 hours to figure it out.

This fixes #604. It might also fix #521?

### DIFF
--- a/answerqueue.go
+++ b/answerqueue.go
@@ -24,9 +24,30 @@ type AnswerQueue struct {
 	method   Method
 	draining chan struct{} // closed while exiting queueing state
 
-	mu    sync.Mutex
-	q     []qent // non-nil while queueing
-	bases []base // set when drain starts. len(bases) >= 1
+	mu sync.Mutex
+	// The entries of the queue. The queue refers to method calls
+	// after an earlier call finishes. But not all entries need to refer
+	// to the same call. Entries can refer to *earlier* entries in
+	// the queue as their "basis". The basis for the initial fulfilled
+	// value is always 0. Take for instance the call chain:
+	//
+	// A() -> a
+	// a.B() -> b
+	// b.C() -> c
+	// a.D() -> d
+	//
+	// Call A is pipelined to B which is pipelined to C. This would
+	// produce a chain of 2 qents with the bases 0 and 1, where 0 is
+	// the fulfilled result of A, and 1 is the result of B. If we add
+	// in a qent to the end for call D pipelined from A, it would have
+	// a basis 0 like with call B.
+	//
+	// This field is set to nil when draining starts.
+	q []qent
+	// Message targets derived from applying a qent. Its length always
+	// at least 1, since the 0 index is used for the initial fulfilled result.
+	// This is set when draining starts.
+	bases []base
 }
 
 // qent is a single entry in an AnswerQueue.
@@ -77,7 +98,10 @@ func (aq *AnswerQueue) Fulfill(ptr Ptr) {
 	for i := range q {
 		ent := &q[i]
 		recv := aq.bases[ent.basis].recv
-		recv(ent.ctx, ent.path, ent.Recv)
+		pcall := recv(ent.ctx, ent.path, ent.Recv)
+		// The basis for our result will always be our index in the queue + 1
+		// since 0 is used for the initial fulfilled result.
+		aq.bases[i+1].recv = pcall.PipelineRecv
 	}
 }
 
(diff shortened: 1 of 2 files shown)


lthibault:
Oh, nice find!  Yeah, seems like it could be the fix for both of those issues. I'll keep my eyes peeled.

Do you see an easy way to add a regression test?

lthibault (dismissed):

ObsidianMinor:
I'm working on that right now, I think I can add something to `level1_test` easy enough?

lthibault:
Sounds perfect! Thank you!

lthibault:
For clarity:  I'll wait for you to push tests to this PR and merge them in together.

Maybe we want to cut a release once this is merged?

ObsidianMinor:
Ended up writing a separate file. A release would be good since this seems like a critical issue in a few cases.

lthibault (approved):
Brilliant! Merging!

Planning on tagging v3.1.0-alpha.1, since @matheusd's prior work introduces a breaking change.

PR #607, opened 2025-10-21
Fix address-instability in MultiSegmentArena by storing *Segment and validating via (id, arena storage)

### LINKED ISSUE #606: MultiSegmentArena: preferred segment is not part of the arena
Hi. We noticed, after updating to the latest go-capnp version (v3.1.0-alpha.1), that we have some test code triggering a "new struct: preferred segment is not part of the arena" error.

Some (very simplified) code that reproduces the issue is:

```
struct KV {
    key @0 :Int32;
    value @1 :Data;
}
```

```go
func TestMultiSegment(t *testing.T) {
	_, seg, err := capnp.NewMessage(capnp.MultiSegment(nil))
	if err != nil {
		t.Fatal(err)
	}
	for i := 0; i < 100; i++ {
		kv, err := NewKV(seg)
		if err != nil {
			t.Fatalf("NewKV(%d): %v", i, err)
		}
		kv.SetKey(int32(i))
		err = kv.SetValue(bytes.Repeat([]byte{byte(i)}, 1024)) // large enough to fill the current segment
		if err != nil {
			t.Fatalf("SetValue(%d): %v", i, err)
		}
                // in a real application, KV would be set to some field of a root message
	}
}
```

which results in
```
NewKV(5): new struct: preferred segment is not part of the arena
```

I think the issue might be here:
https://github.com/capnproto/go-capnp/blob/0b34935959cb2d6d2e99b5814e9dd733d2c6b105/arena.go#L303

We keep passing in the same segment pointer, however the MultiSegmentArena compares it to the address of the segments in the `msa.segs` slice, whose underlying array can however be reallocated in the meantime (which is why the error only happens on the 6th iteration).

If my understanding is correct, and the issue is with MultiSegmentArena and not with the way we are using the library, a solution would be turning that slice of segment to a slice of pointers to segments. I'm happy to open a PR in that case.

Thanks in advance!

saikocat:
This attempts to fix #606 - `MultiSegmentArena`: `preferred segment is not part of the arena`

* Replace `[]Segment` with `[]*Segment` in MultiSegmentArena to avoid address-instability on append.
* Make `Segment(id)` return the stable pointer from `msa.segs` (no `&msa.segs[id]`).
* In `Allocate`, validate the preferred segment in O(1) by checking `seg.id` is in-range and `msa.segs[seg.id] == seg`; remove the linear pointer scan.
* Keep `msa.segs` length in sync with the byte-storage (`msa.segs[i].data`).
* Ensure `Release()` zeroises and returns buffers, then clears pointers safely.

### DIFF
--- a/arena.go
+++ b/arena.go
@@ -157,7 +157,7 @@ func (ssa *SingleSegmentArena) Release() {
 // buffers, allocating new buffers of exponentially-increasing size when
 // full. This avoids the potentially-expensive slice copying of SingleSegment.
 type MultiSegmentArena struct {
-	segs []Segment
+	segs []*Segment
 
 	// rawData is set when the individual segments were all demuxed from
 	// the passed raw data slice.
@@ -204,12 +204,16 @@ func (msa *MultiSegmentArena) Release() {
 	msa.rawData = nil
 
 	for i := range msa.segs {
+		if msa.segs[i] == nil {
+			continue
+		}
 		if msa.bp != nil {
 			zeroSlice(msa.segs[i].data)
 			msa.bp.Put(msa.segs[i].data)
 		}
 		msa.segs[i].data = nil
 		msa.segs[i].BindTo(nil)
+		msa.segs[i] = nil
 	}
 
 	if msa.segs != nil {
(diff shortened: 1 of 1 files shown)


lthibault (approved):
LGTM.  Approving once it clears CI.

lthibault:
Do I need to cut a release?

saikocat:
Yes, I think so. This is due to previous version is buggy with MultiSegment.

